PR #8, opened 2015-08-16
Adding the Reactions endpoint to the client

jakedahn:
This PR introduces the `Reactions` endpoint to the client. I didn't see any examples of other endpoint tests so I'm not sure if this is a finished PR quite yet.

Please let me know if anything else is needed to get this merged, I'm happy to make updates.


dblock:
Add a line into the CHANGELOG please? When and where does slack announce these new things? We should have some kind of trigger on their docs - maybe a bot? ;)

dblock:
Since these are auto generated I didn't feel they need tests. A good idea could be to auto generate tests (obviously beyond the scope of this PR). And maybe have a test that checks whether this lib is out of date with slack docs.

dblock:
Bump @jakedahn.

jakedahn:
@dblock oops sorry, I haven't had a chance to look at things since I submitted the PR.

I will take a look at things later tonight and push some changes :)

jakedahn:
I've added a change to the changelog.

Also I didn't notice that the API was auto generated at first glance and just copy-pasted things until I got a result that worked. I did a little digging to see how the auto generation worked and found that the way things work may be depending on an approach that appears to be no longer maintained.

https://github.com/slackhq/slack-api-docs has not been updated for several months, while the public api docs website has been updated several times. I've opened up a support ticket with @slackhq to see if they plan on updating/maintaining the `slack-api-docs` repo publicly, or if there is another approach for getting at the newly updated json schema files that are used for the client generation. I'll report back when I get more details.

dblock:
I noticed that too. I also told @busterbenson at Slack, but this hasn't had any traction. I think we need a plan B, and start maintaining our own version or fork for that or maybe extract this content from their website?

dblock:
@jakedahn Slack said they will not be updating that repo until they hire someone. We don't have to wait. I've pulled JSON files from https://github.com/slackhq/slack-api-docs into https://github.com/dblock/slack-api-ref, which I intend to maintain and kept JSON files only. Could you please PR this method there in JSON and update the client here based on that using the Rake task (I've updated it to pull from the new repo already)?

jakedahn:
@dblock sure thing! I will likely not get to it until this weekend when I'm hacking on related stuff again.

jakedahn:
@dblock I've opened up a PR here: https://github.com/dblock/slack-api-ref/pull/3

I've updated this PR by running `rake slack:web:api:update` against a local, modified copy of `slack-api-ref` (the same sha that is proposed in the other repo). I've also confirmed that the updated client works great with the app that I'm building :tada:

One thing I don't know how to address is the line-length linting issue that showed up on travis. In the slack-api-ref PR I used description strings taken from the public website, which are verbose strings, and the new url for the json specs exceed the max line length as well.

dblock:
Just run `rubocop --auto-gen-config`.

jakedahn:
@dblock I've upgraded the `slack-api-ref` submodule sha,  and squashed down to a single commit, I think this one is ready for a merge

dblock:
Great, thanks!

PR #11, opened 2015-08-20
Chat postMessage argument error only if both attachments and text are missing

RoryDH:
The slack chat http api allows you to send no text field if you do provide attachments. This change will stop the local validation from blocking that use case.

### DIFF
--- a/lib/slack/web/api/endpoints/chat.rb
+++ b/lib/slack/web/api/endpoints/chat.rb
@@ -49,7 +49,7 @@ module Slack
           # @see https://github.com/slackhq/slack-api-docs/blob/master/methods/chat.postMessage.json
           def chat_postMessage(options = {})
             throw ArgumentError.new('Required arguments :channel missing') if options[:channel].nil?
-            throw ArgumentError.new('Required arguments :text missing') if options[:text].nil?
+            throw ArgumentError.new('Required arguments :text and :attachments missing') if options[:text].nil? && options[:attachments].nil?
             post('chat.postMessage', options)
           end
 


dblock:
This file is generated [by a rake task](https://github.com/dblock/slack-ruby-client/blob/master/lib/slack/web/api/tasks/generate.rake#L14) and will be overwritten next time Slack updates their web API. So we need a solution that fixes this during generation.

dblock:
I think the problem is that https://github.com/slackhq/slack-api-docs/blob/master/methods/chat.postMessage.json says that `text` is required, which you're saying is not true if attachments are there. You should at the very least open a bug in https://github.com/slackhq/slack-api-docs.

RoryDH:
Ah I see, thanks for the explanation, I'll do that now.

dblock:
I am open for a way to fix this here with a work-around. Generally slack API docs are not complete or often miss things. Maybe we can have a patches folder that applies diffs to the generated files or something like that? I suspect it will take a long time for Slack to figure out how to express "one of these two parameters is required" in JSON :)

RoryDH:
All true.

I'll open the issue anyway and hopefully get around building some kind of a system like what you described.

dblock:
Already opened https://github.com/dblock/slack-ruby-client/issues/12 ;)

RoryDH:
Thanks! I'll open one their api-docs too though

RoryDH:
slackhq/slack-api-docs#41

dblock:
@RoryDH Slack said they will not be updating that repo until they hire someone. We don't have to wait. I've pulled JSON files from https://github.com/slackhq/slack-api-docs into https://github.com/dblock/slack-api-ref, which I intend to maintain and kept JSON files only. 

Could you please help with this issue? Maybe we can update the JSON schema to support "one of these parameters is required" and fix slack-ruby-client?

dblock:
I've opened a ticket with Slack on fixing the documentation, https://artsy.slack.com/help/requests/610828.

dblock:
Implemented as patches in https://github.com/dblock/slack-ruby-client/commit/84416a94330c6a05c24828e1604a43d4e5556b03.

Issue #5, opened 2015-08-05
Add Celluloid support next to EventMachine


dblock:
See https://github.com/abstractive/slack-ruby-gem/commit/41a45ba11a11c087b501654a159e72f27526dd9d.

digitalextremist:
I'll add Celluloid support, as I have in the other gem. Two things though. I'd actually use `Reel::IO` ( celluloid/reel-io#2 ) and I'd also use [this http gem](https://github.com/httprb/http) instead.

dblock:
@digitalextremist are you saying we should replace Faraday for httprb? I would be open to it, but I am not sure it matters for slack-ruby-client given that HTTP communication is hardly the overhead.

digitalextremist:
Correct, it is unrelated to the `WebSocket` aspect, so I will post a second ticket for this piece. Why I mention it is that this `http` gem is compatible with `Celluloid::IO` which makes it evented, and in my uses, I do have many concurrent requests taking place at once in many cases.

mikz:
I'm really interested in this. Is there a way I could help?

dblock:
Please contribute @mikz!

mikz:
I'd like not to clash on some ongoing work. I plan to make a prototype using just celluloid-io and websocket-driver without using reel-io.

mikz:
@dblock @digitalextremist I did a prototype. See the code in https://github.com/dblock/slack-ruby-client/commit/ced303d099334a1a7dd191da827e3676487a6228. It successfully connected to slack and was answering with celluloid.

The celluloid integration is probably not very correct. It crashes on Ctrl-C. 

What I'd like to do next:

* cover it with tests that work for both concurrency options
* refactor the socket / client so it is more understandable

dblock:
This is very good. I actually think this whole code should be a gem that abstracts various ways of doing concurrency.

mikz:
@dblock opened #15 for discussion

tuned a bit the original prototype so now it detects what you have available
tried both realtime examples and they work

it is possible that some cases are broken, like closing connection or processing some messages
but rather to show it early than go the wrong path

dblock:
Closing via https://github.com/dblock/slack-ruby-client/pull/26. Please try Celluloid support on HEAD.

PR #15, opened 2015-09-15
experimental support for both eventmachine and celluloid

mikz:
* uses autoload to lazy load dependencies on access
* slack/real_time/celluloid requires celluloid
* slack/real_time/eventmachine requires eventmachine
* moves the gem dependencies to development
* configure concurrency through .configure

implements #5


dblock:
This is awesome work. Read through the code, all looks good.

* Fix build, `rubocop -a`, `rubocop --auto-gen-config`.
* Document in README and add to CHANGELOG.
* Add a test for `NoConcurrencyError` being raised.

I think it will be good to merge after.

lib/slack/real_time/celluloid.rb
```diff
@@ -0,0 +1,76 @@
...
+
+          async.run_loop
+        end
+
+        def run_loop
+          loop { read } if socket
+        end
+
+        def read
+          buffer = socket.readpartial(4096)
```
dblock: Extract as constant.

lib/slack/real_time/celluloid.rb
```diff
@@ -0,0 +1,76 @@
...
+
+        protected
+
+        def connected?
+          !!@connected
+        end
+
+        def build_socket
+          socket = TCPSocket.new(addr, port)
+          socket = SSLSocket.new(socket, ssl_context) if secure?
```
dblock: You're making a `TCPSocket` and an `SSLSocket` this should be `socket = secure? ? ...`.

lib/slack/real_time/celluloid.rb
```diff
@@ -0,0 +1,76 @@
...
+          socket
+        end
+
+        def ssl_context
```
dblock: Maybe `make_ssl_context` or `build_ssl_context`, since it's not memoized, just to be consistent? No strong feelings.

lib/slack/real_time/config.rb
```diff
@@ -15,6 +18,23 @@ def reset
         self.websocket_ping = 30
         self.websocket_proxy = nil
         self.token = nil
+        self.concurrency = method(:detect_concurrency)
+      end
+
+      def concurrency
+        (val = @concurrency).respond_to?(:call) ? val.call : val
```
dblock: Trying to be too clever. This actually doesn't save any cycles and is probably even slower than 

```ruby
@concurrency.respond_to?(:call) ? @concurrency.call : @concurrency
```

lib/slack/real_time/config.rb
```diff
@@ -15,6 +18,23 @@ def reset
...
+      def detect_concurrency
+        [:Eventmachine, :Celluloid].each do |concurrency|
+          begin
+            return Slack::RealTime.const_get(concurrency)
+          rescue LoadError
+            # could not be loaded, missing dependencies
+          end
+        end
+
+        raise NoConcurrencyError, 'could not find concurrency primitives, add faye-websocket or celluloid-io'
```
dblock: Probably should say "to Gemfile", and definitely needs to be documented.

lib/slack/real_time/eventmachine.rb
```diff
@@ -0,0 +1,27 @@
...
+  module RealTime
+    module Eventmachine
+      class Socket  < Slack::RealTime::Socket
+
+        def self.close
+          ::EM.stop
+        end
+
+        def self.run(*args)
+          EM.run {
```
dblock: `EM.run do ...`

lib/slack/real_time/socket.rb
```diff
@@ -4,39 +4,65 @@ class Socket
...
+
       def initialize(url, options = {})
         @url = url
         @options = options
+        @driver = nil
       end
 
       def send_data(data)
-        @ws.send(data) if @ws
+        driver.text(data) if driver
```
dblock: Maybe `text` is incorrect here. Since in many cases this is allowed to be random binary data. Change to `send_data`?

lib/slack/real_time/socket.rb
```diff
@@ -4,39 +4,65 @@ class Socket
...
+            URI::HTTPS::DEFAULT_PORT
+          when 'ws', 'http'.freeze
+            URI::HTTP::DEFAULT_PORT
+          else
+            uri.port
+        end
+      end
+
+      def connect
+        raise "Expected #{self.class} to implement #{__method__}"
```
dblock: This is missing a period at the end.

Also NoConcurrencyError should be a full phrase then with a period at the end.

spec/slack/real_time/celluloid_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+        expect(socket.url).to eq url
+      end
+    end
+
+    describe '#connect!' do
+      before do
+        allow(ws).to receive(:on).with(:close)
+      end
+
+      xit 'connects' do
```
dblock: These should work, no?

lib/slack/real_time/celluloid.rb
mikz: the `SSLSocket` has to wrap the `TCPSocket` so it could be:

```ruby
secure ? SSLSocket.new(socket, ssl_context) : socket
```

lib/slack/real_time/celluloid.rb
mikz: Makes total sense.

lib/slack/real_time/config.rb
mikz: I did not do it for speed, but for being dry. I really dislike repeating long instance variable names because they just return nil if you mistype them.

lib/slack/real_time/config.rb
mikz: Yep. I was thinking about creating meta dependency. Like `slack-ruby-client-eventmachine` and `slack-ruby-client-celluloid` which would depend on proper `slack-ruby-client` and faye/celluloid.

lib/slack/real_time/socket.rb
mikz: Yep text is not right, but send_data is implemented just in faye-websocket. Will need to extract it.

spec/slack/real_time/celluloid_spec.rb
mikz: Problem is that it actually starts the loop so it hangs the test.

mikz:
@dblock do you have some real app to try this out? 

I suspect it broken the disconnect logic and maybe some other parts. 

I was thinking about doing some integration test? So anyone could try it out with their own slack token.

lib/slack/real_time/celluloid.rb
dblock: Bah I was not awake yet :)

lib/slack/real_time/config.rb
dblock: Fair.

lib/slack/real_time/config.rb
dblock: The meta dependency is a pretty neat idea. But it would require 2 releases every time and I am not sure it's such a great idea.

Also, btw, maybe if you have *both* you have to choose explicitly? There should probably be tests for that.

lib/slack/real_time/socket.rb
dblock: Yep and that's actually something I use frequently, for the RTM client you need to be able to send back data all the time.

spec/slack/real_time/celluloid_spec.rb
dblock: Stub away the loopy parts. Like `EM.start do ...` can be `allow(EM).to receive(:start).and_yield`

dblock: Another option is to create a dummy driver implementation for tests only.

dblock:
You could try with https://github.com/dblock/slack-gamebot for example. Just get yourself a test Slack team.

dblock:
An integration test would be :+1:.

mikz:
@dblock pushed some style fixes and  WIP commit cba0160 which has working integration test
the client basically talks to itself to verify it works :)

dblock:
I think this is close, bring it to a mergeable state.

dblock:
@mikz Want to finish this? Need help?

mikz:
I just need some time. Will take a stab  in it this week. The next surely no as I'll be traveling.
Would be great to try it with real world app. I'm trying it with the game not, but still the more the better.

dblock:
I continued this and made a bunch of spec fixes and such, lets close this in favor of https://github.com/dblock/slack-ruby-client/pull/26. Right now my Celluloid integration tests hang, but everything else seems to be fine. I could use some help with that one.

mikz:
That is amazing! Thanks so much. Will try to incorportate into  real project soon.

PR #26, opened 2015-11-13
Support both EventMachine and Celluloid concurrency.

dblock:
This is a continuation of https://github.com/dblock/slack-ruby-client/pull/15.

I replaced the integration test's Fiber implementation with a thread and a queue and that now succeeds for both Celluloid and EventMachine.


dblock:
... and merged, it's working well enough for other people to try, will play around with some bots and Celluloid

PR #39, opened 2015-12-28
Update README.md

jasonbain:
Added section and example code to use the RealTime client to upload a file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,54 @@ See a fullly working example in [examples/hi_real_time_and_web](examples/hi_real
 
 ![](examples/hi_real_time_and_web/hi.gif)
 
+#### Upload a file
+The RealTime client can be used to upload a file using the [Slack files.upload API](https://api.slack.com/methods/files.upload).  Here is a quick example of uploading a file to a Slack Channel:
+
+```ruby
+require 'slack-ruby-client'
+
+path_to_file = "/path/to/sample/file.txt"
+filetype = 'text'
+
+
+def send_file(client,data,path_to_file,filetype)
+  puts "sending file #{path_to_file} to #{data['user']}"
+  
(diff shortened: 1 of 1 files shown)


dblock:
This is good for a blog post, but not great for documentation.

* You're using the web client that happens to be accessible via the RealTime client (because it was used to do `rtm.start`), so it's incorrect to state that this is done with a RealTime client.
* There's a ton of boilerplate unrelated to actually uploading a file.

I'll give this a shot based on your example.

dblock:
Closing in favor of https://github.com/dblock/slack-ruby-client/commit/c7dd2b5be8c0a60aec37cd8836527c3fec1071ad.

jasonbain:
Daniel,

I have been using this in my code using a bot that listens to our channels.  It works.  If you have recommendations on improvement I would love to hear it.  I am always looking to improve my code.  Right now I am calling a function to send logs after our DevOpsBot takes action...

```ruby
def send_log(client,data,file,message)
  puts "sending file to #{data['user']}"

  client.web_client.files_upload(channels: data['channel'], as_user: true,
                                 initial_comment: "<@#{data['user']}>, #{message}",
                                 file: Faraday::UploadIO.new(file,'text'),
                                 title: File.basename(file),
                                 filename: File.basename(file)
                                )
end
```

I am sure there is a better way to do this.  Having an example would have saved me hours of trial and error.  However, pouring over the Slack API docs was very educational.  It is a bummer there isn't an authoritative example on the proper way for a bot integration to do this kind of action.

dblock:
This looks totally fine. The only thing I would do is to queue the sending of the file because this potentially can take a long time and is likely blocking the message queue of the bot. That would require something more complicated in your bot, like a Redis queue and a worker.

PR #40, opened 2016-01-06
Client stop method fails as the connected? method is a protected in Slack::RealTime::Concurrency::Celluloid::Socket class

kandadaboggu:
Made the `connected?` method public in [Slack::RealTime::Concurrency::Celluloid::Socket](https://github.com/workato/slack-ruby-client/blob/5b83b24cb6591c09b26c09bb29ce47dd9d637702/lib/slack/real_time/concurrency/celluloid.rb#L35)

### DIFF
--- a/lib/slack/real_time/concurrency/celluloid.rb
+++ b/lib/slack/real_time/concurrency/celluloid.rb
@@ -32,6 +32,10 @@ module Slack
             future.run_loop
           end
 
+          def connected?
+            !@connected.nil?
+          end
+
           def run_loop
             loop { read } if socket
           rescue EOFError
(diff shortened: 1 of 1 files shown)


dblock:
Good point. This needs a test, please and an entry in https://github.com/dblock/slack-ruby-client/blob/master/CHANGELOG.md, please.

dblock:
Bump @kandadaboggu I tried to reproduce this, but I can't, the protected method is called from the same namespace and seems to work just fine as is.

dblock:
Closed via https://github.com/dblock/slack-ruby-client/pull/49.

kandadaboggu:
@dblock Sorry didn't see the message in time. It looks like #49 has added the requisite tests.

PR #50, opened 2016-01-22
[wip] logging socket data

mikz:
it is quite useful when developing, debugging, ...
will try to add more places and make the logger configurable

wdyt?

### DIFF
--- a/lib/slack/real_time/concurrency/eventmachine.rb
+++ b/lib/slack/real_time/concurrency/eventmachine.rb
@@ -15,11 +15,11 @@ module Slack
           end
 
           def send_data(message)
+            logger.debug('#send_data') { message }
             driver.send(message)
           end
 
           protected
-
           # @return [Thread]
           def ensure_reactor_running
             return if EventMachine.reactor_running?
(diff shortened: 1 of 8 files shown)


dblock:
I like it.

I wouldn't log the method names by interpolating strings, different people want different styles of what is being logged, so I would just do `logger.debug buffer` and not try to impose a format on anyone.

mikz:
@dblock then I'd go with `logger.debug('[slack#write]') { buffer }`. You can ignore the prefix by setting log format. 
Some of the buffers are binary, and without knowing from where they come, it is madness.

dblock:
:+1: 

I think logger then should become part of `Slack::RealTime::Socket`, and so on and so forth and be documented.

mikz:
@dblock pushed more stuff. It is WIP. 
Two parts I'm not happy with:

* eventmachine driver instrumentation
* the code to print events

The later could be fixed, but the eventmachine don't think so.

lib/slack/real_time/concurrency/eventmachine.rb
```diff
@@ -30,7 +30,23 @@ def ensure_reactor_running
           end
 
           def connect
-            @driver = ::Faye::WebSocket::Client.new(url, nil, options)
```
dblock: Did you try inheriting from `Faye::WebSocket::Client` and overriding `parse` and `write` methods? And instantiate the child class here?

dblock:
I see what you mean wrt Eventmachine, see my suggestion above.

dblock:
I want to release 0.6.0 soon, maybe you can find a bit of time to finish this? Can I help?

mikz:
@dblock will spend some time on it on wednesday and thursday.

dblock:
Yeah, I'll do a release before that.

dblock:
I am still looking forward to this @mikz!

mikz:
@dblock whenever I'll do another round of bot development, I'll definitely do this.

dblock:
Closing in favor of https://github.com/dblock/slack-ruby-client/pull/67. Fixed tests, made some improvements, still a few TODOs remaining.

PR #53, opened 2016-01-26
#9 Auto-generate tests for client APIs

accessd:
Tried to implement #9.
May be need something else :)


dblock:
This is great. I am going to merge. 

Now, what I had in mind goes much further, I am going to describe it in #9.

PR #62, opened 2016-02-23
fix url

harlantwood:
There are may be more like this, just happened to notice this one...


dblock:
Thanks. However this file is auto-generated. The fix should be in one of the https://github.com/dblock/slack-ruby-client/tree/master/lib/slack/web/api/templates files and the rake task re-run, please.

Issue #42, opened 2016-01-08
Ruby 2.3.0 suport


dblock:
Needs a release of celluloid-io, https://github.com/celluloid/celluloid-io/pull/163.

harlantwood:
Looks like that's done -- https://rubygems.org/gems/celluloid-io/versions/0.17.3

Any other known issues with Ruby 2.3.0?

dblock:
Closed via https://github.com/dblock/slack-ruby-client/pull/65.

PR #65, opened 2016-02-26
Add Ruby 2.3.0 to Travis

harlantwood:
- Tests pass under ruby 2.3: https://travis-ci.org/harlantwood/slack-ruby-client/builds/112131523
- For whatever reason the string "2.3" doesn't work, but "2.3.0" does (rvm issue perhaps)

See also #42


dblock:
:+1:

PR #67, opened 2016-02-29
Make logger configurable and log socket data.

dblock:
This is a replacement for https://github.com/dblock/slack-ruby-client/pull/50.


dblock:
@mikz FYI I finished what you started in #50 here.

lib/slack/real_time/client.rb
```diff
@@ -125,6 +127,7 @@ def socket_class
 
       def send_json(data)
         fail ClientNotStartedError unless started?
+        logger.debug("#{self.class.name}#send_data") { data }
```
mikz: shouldn't this be `#send_json` ? maybe using `"#{self.class}#${__method__}"` would make it automatic.

lib/slack/real_time/concurrency/eventmachine.rb
```diff
@@ -5,6 +5,26 @@ module Slack
...
+          attr_reader :logger
+          protected :logger
+
+          def initialize(url, protocols = nil, options = {})
+            super
+            @logger = options.delete(:logger) || Slack::RealTime::Config.logger || Slack::Config.logger
+          end
+
+          def parse(data)
+            logger.debug("#{self.class.name}#read") { data }
```
mikz: should this one be called parse?

mikz:
This looks great! Good job :+1:

lib/slack/real_time/client.rb
dblock: :+1: 

Generally I wonder about logging data here, it's a binary blob most of the time. Thoughts?

PR #69, opened 2016-03-09
Add attachments support for Slack::Web::Api::Endpoints::Chat.chat_update

nicka:


dblock:
Unfortunately this is not as easy :) This file is generated by a Rake task from documentation. If slack has documented this, the first step is to update https://github.com/dblock/slack-api-ref. Then in this project you can run `rake slack:api:update`. Everything on top is a patch, you can see how it's done for `chat_postMessage` in https://github.com/dblock/slack-ruby-client/tree/master/lib/slack/web/api/patches. The specs should be good though.

Also will need a CHANGELOG entry, please.

nicka:
@dblock Sure thing, will fix it. Do you also want me to do a version bump? Or something like `Next version` within the CHANGELOG.md?

nicka:
@dblock Added the patch :+1: and updated the `CHANGELOG.md`.

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ### 0.7.1 (Next)
 
-* Your contribution here.
```
dblock: Put this back please?

lib/slack/web/api/endpoints/chat.rb
```diff
@@ -82,8 +82,14 @@ def chat_postMessage(options = {})
...
             throw ArgumentError.new('Required arguments :channel missing') if options[:channel].nil?
-            throw ArgumentError.new('Required arguments :text missing') if options[:text].nil?
+            throw ArgumentError.new('Required arguments :text or :attachments missing') if options[:text].nil? && options[:attachments].nil?
             options = options.merge(channel: channels_id(options)['channel']['id']) if options[:channel]
+            # attachments must be passed as an encoded JSON string
+            if options.key?(:attachments)
+              attachments = options[:attachments]
+              attachments = JSON.dump(attachments) unless attachments.is_a?(String)
+              options = options.merge(attachments: attachments)
+            end
```
dblock: I think if we're going to do this we should do it in `chat_postMessage` too. Maybe simplify the code:

```ruby
attachments = JSON.dump(options[:attachments]) if options.key?(:attachments) && ! options[:attachments].is_a?(String)
```

dblock: Oh we already do that in `chat_postMessage`, my bad. You can maybe make the above change in a future commit (but feel free to do it here too).

dblock:
Would have merged minus my removed line in CHANGELOG. Looking forward to an `--amend`. Thanks!

nicka:
@dblock Done!

dblock:
Merged.

PR #70, opened 2016-03-11
update to support listing the users in a usergroup and updating the users in the usergroup

mooremo:


dblock:
Unfortunately this is not as easy :) This file is generated by a Rake task from documentation. If slack has documented this, the first step is to update https://github.com/dblock/slack-api-ref. Then in this project you can run `rake slack:api:update`.

mooremo:
I see. I'll close this PR and open another after I've done that.

PR #73, opened 2016-03-14
add a 'closed' event

rkadyb:
I ended up needing an 'after-close' event for a project I'm working on, figured I'd share this in case it's worthwhile for anyone else.

### Reasoning:
The existing `:close`event is fired BEFORE the websocket connection is actually closed, which makes sense if you want to send a message / use the API immediately before disconnect. This isn't as helpful if you want to do something not related to the API AFTER the disconnect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### 0.7.1 (Next)
 
 * Your contribution here.
+* [#73](https://github.com/dblock/slack-ruby-client/issues/73): Add a `closed` event - [@rkadyb](https://github.com/rkadyb).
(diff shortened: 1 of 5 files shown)


dblock:
Looks good. Needs CHANGELOG, tests and a README update please. Maybe call it `:closed` instead?

dblock:
Bump, maybe you want to finish this one @rkadyb ?

rkadyb:
Sorry for the delay, will finish this one up shortly!

rkadyb:
I cleaned this PR up a bit, please let me know what you think.

dblock:
I like it! Squash it and I'll merge?

rkadyb:
Squashed, thanks!

dblock:
I merged. Before the next version ships, maybe we want a proper callback mechanism added here where you can do things before, after and around something, ActiveSupport style. Maybe you might want to take a look @rkadyb.

dblock:
Opened https://github.com/dblock/slack-ruby-client/issues/76.

PR #74, opened 2016-03-16
Timeout

technicalpickles:
This adds support for a timeout as described in https://github.com/dblock/slack-ruby-client/issues/72

Implementation is inspired by https://github.com/getsentry/raven-ruby/pull/54/files

### DIFF
--- a/lib/slack/config.rb
+++ b/lib/slack/config.rb
@@ -4,10 +4,14 @@ module Slack
 
     attr_accessor :token
     attr_accessor :logger
+    attr_accessor :timeout
+    attr_accessor :open_timeout
 
     def reset
       self.token = nil
       self.logger = nil
+      self.timeout = nil
+      self.open_timeout = nil
     end
 
     reset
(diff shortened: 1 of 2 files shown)


dblock:
This only affects the web client, so should be part of Slack::Web::Config.  It needs tests, README and CHANGELOG entry.

But I'd really rather not add this, but expose the connection during construction similar to how https://github.com/codegram/hyperclient does it, this way we don't have to add every option to the config. Or at least having a hash for connection options in the config so that you can throw whatever you want in there.

technicalpickles:
I was using this branch to address timeouts I was getting from Slack, but I couldn't actually set it high enough to finish without Slack resetting the connection. I'm already working with Slack to track down what the problem is, but I'm not sure if it's a problem with the timeout as there is with file processing on their end.

dblock:
Bump, maybe you want to try to turn this into something merge-able @technicalpickles ?

asheynkman:
any plans to push this fix into release ? so timeout can be controlled for bigger uploads and downloads ?

dblock:
No "plans", someone needs to finish this PR, see my comments above.

technicalpickles:
Unfortunately it has been long enough that I don't even remember what I was using this for, and the issue I linked to is a 404 😦 

I'm going to close, but will leave the branch up if anyone wants to build on it.

asheynkman:
hi Guys ,
 sorry been little busy did not get to this yet, I need to add respect and documentation and will created pull request. We have increase faraday timeout and that is resolve download initial connection metadata.

dblock:
I've PRed https://github.com/slack-ruby/slack-ruby-client/pull/135 that introduces timeout options with tests and READMe.

PR #82, opened 2016-04-29
Fix usergroups.users.{list,update} and files.commnets.{add,edit,delete} APIs

masatomo:
Some Slack APIs has 2 dots like usergroups.users.{list,update} / files.commnets.{add,edit,delete}
These APIs didn't work and this patch fixes it.


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ### 0.7.1 (Next)
 
 * Your contribution here.
+* [#82](https://github.com/dblock/slack-ruby-client/pull/82): Fix usergroups.users.{list,update} and files.comments.{add,edit,delete} APIs
```
dblock: Make this look like the other entries please, quote the methods and add your `- name.` at the end.

bin/commands/usergroups_users.rb
```diff
@@ -0,0 +1,25 @@
+# This file was auto-generated by lib/tasks/web.rake
+
+desc 'Usergroups.users methods.'
```
dblock: This looks strange. These should be subcommands of usergroups, see how it moved from the other file?

dblock:
Do I understand correctly that the implementation here of `files_comments` is actually not an API, the API is `files_comments_edit` and such? That makes sense, so the only problem in this PR I think is the command line implementation where you want `files comments edit` kind of API.

masatomo:
> Do I understand correctly that the implementation here of files_comments is actually not an API, the API is files_comments_edit and such?

Right. refs. https://api.slack.com/methods/files.comments.edit  

Also, I've updated everything you pointed out.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ### 0.7.1 (Next)
 
 * Your contribution here.
+* [#82](https://github.com/dblock/slack-ruby-client/pull/82): Fix `usergroups.users.{list,update}` and `files.comments.{add,edit,delete}` APIs - [@masatomo](https://github.com/masatomo)
```
dblock: Nitpick: Missing period at the end.

bin/commands/files_comments.rb
```diff
@@ -0,0 +1,35 @@
...
+  g.long_desc %( Edit an existing comment on a file. Only the user who created a comment may make edits. Teams may configure a limited time window during which file comment edits are allowed. )
+  g.command 'edit' do |c|
+    c.flag 'file', desc: 'File containing the comment to edit.'
+    c.flag 'id', desc: 'The comment to edit.'
+    c.flag 'comment', desc: 'Text of the comment to edit.'
+    c.action do |_global_options, options, _args|
+      puts JSON.dump($client.files_comments_edit(options))
+    end
+  end
+end
```
dblock: Can we make this `files comments edit` instead of `files_comments edit`? These commands can nest multiple levels.

dblock: Actually I'll take it back. Ignore me this is right.

dblock:
Please check the build, you have to fix `rubocop` at the very least, run `rubocop -a`, then `rubocop --auto-gen-config` for everything else. Squash your commits. Thanks, this is close to be merged.

masatomo:
ok. but seems `robocop -a` updates many files which I didn't touch for this pull-request. do you still want me to run it?

dblock:
It shouldn't but yes, @masatomo, just make the build pass and then we shall see :)

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ### 0.7.1 (Next)
 
 * Your contribution here.
+* [#82](https://github.com/dblock/slack-ruby-client/pull/82): Fix `usergroups.users.{list,update}` and `files.comments.{add,edit,delete}` APIs. - [@masatomo](https://github.com/masatomo)
```
dblock: The period is in the wrong place, should be at the end of the line :)

masatomo:
seems rubocop verison  in trivis-ci is older than mine what do you want me to do?

dblock:
It uses what's inside Gemfile.lock, you probably have a global version and would need to run this with `bundle install`. I'll take care of this, no worries.

dblock:
Merged via https://github.com/dblock/slack-ruby-client/commit/90e69ad1f7df3d40956c44875df72383d3fe04f8. Thank you.

masatomo:
ok, thanks!

Issue #92, opened 2016-05-20
Celluloid::IO support handles concurrency incorrectly


benzrf:
The advantage of `Celluloid::IO` is that it gives you the ability to make IO calls which suspend the current task but do not block the actor as a whole. The realtime client doesn't make proper use of this, though - after reading from the socket, it immediately handles the input synchronously within the same task, so the input loop will not proceed to the next iteration until the handler is done. The net result is that the program can only handle one incoming message at once.

benzrf:
Here's an example:

```ruby
require 'celluloid/current'
require 'slack-ruby-bot'

class TestBot < SlackRubyBot::Bot
  command 'expensive' do |client, data, match|
    Celluloid::Actor.current.sleep 2
    client.say(text: 'Done with expensive', channel: data.channel)
  end

  command 'cheap' do |client, data, match|
    client.say(text: 'Done with cheap', channel: data.channel)
  end
end

TestBot.run
```

https://gfycat.com/LikableLavishAtlanticridleyturtle

dblock:
Thanks. You're right, would really appreciate a PR that fixes this. There're other celluloid-io questions like #84 and #90.

benzrf:
I did manage to make a small change that gets my example bot working, but I'm not really sure what repercussions the change would have, or if other parts might need changing. I'll attach it.

benzrf:
See #93.

benzrf:
Thanks for the merge :)

PR #93, opened 2016-05-20
Handle received data asynchronously

benzrf:
See #92.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,4 +25,5 @@ git:
   submodules: false
 
 before_install:
+  - gem update bundler
   - gem update --system
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,3 +23,6 @@ matrix:
 
 git:
   submodules: false
+
+before_install:
+  - gem update --system
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### 0.7.4 (Next)
 
 * Your contribution here.
+* [#93](https://github.com/dblock/slack-ruby-client/pull/93): Fix: When using Celluloid concurrency, handle input from the TCP socket asynchronously from reading more. - [@benzrf](https://github.com/benzrf).
 
 ### 0.7.3 (5/14/2016)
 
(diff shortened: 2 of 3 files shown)


dblock:
Can you please update CHANGELOG. Could use a test maybe too?

dblock:
The build is breaking for some other reason (something wrong with `vcr`, possibly just need to lock down that one to a previous release - can you please check it out - I am happy to fix if it's too difficult).

benzrf:
What kind of test could I write that wouldn't be extremely nondeterministic?

dblock:
I would just make sure that the function implemented by the bot command is invoked via `async`, so somewhere I'd have something like `expect(...).to receive(:async).and_...`. But if it's too complicated it's fine without it.

benzrf:
Oh, I could probably have the first command sleep indefinitely for the test. I'll try implementing that soon.

benzrf:
OK, it looks like it'd be fairly complicated to write a test. I *might* be able to figure something out, but it'd definitely take a lot of digging and learning about how the library and its tests are set up.

dblock:
This looks good. Could you please try to fix the build or let me know if you can't?

benzrf:
It looks like the issue is with `vcr` - there was a new minor version released today.

benzrf:
https://github.com/vcr/vcr/issues/582

dblock:
Lock vcr to 3.0.1 in this PR?

benzrf:
ok - but it seems like kind of an innapropriate place to do it.

benzrf:
The issue on the vcr repo suggests that this should fix it too

benzrf:
ok nvm, I'll just fix the vcr version.

dblock:
Ha, it's now failing with another issue around bundler. That's https://github.com/bundler/bundler/issues/3558 and the fix is to update bundler:

```ruby
before_install:
   - gem update bundler
```

Please. I just want this build to be green.

benzrf:
Wait - would you prefer that I lock the vcr version, or add bundler update as well?

dblock:
Honestly I don't have a preference. It's just mechanics.

benzrf:
The build is passing

dblock:
Merged, thanks.

PR #98, opened 2016-06-26
add `chat.meMessage`

### LINKED ISSUE #97: chat.rb is missing chat.meMessage
The Slack API supports a method `chat.meMessage` which sends a message like `/me waves`. https://api.slack.com/methods/chat.meMessage

I can't find this method in the `chat.rb` file https://github.com/dblock/slack-ruby-client/blob/master/lib/slack/web/api/endpoints/chat.rb

aaronpk:
closes #97

### DIFF
--- a/lib/slack/web/api/endpoints/chat.rb
+++ b/lib/slack/web/api/endpoints/chat.rb
@@ -63,6 +63,21 @@ module Slack
           end
 
           #
+          # This method sends a me message to a channel from the calling user.
+          #
+          # @option options [channel] :channel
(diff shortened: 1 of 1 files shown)


dblock:
Unfortunately this is not as easy :) This file is generated by a Rake task from documentation. If slack has documented this, the first step is to update https://github.com/dblock/slack-api-ref. Then in this project you can run `rake slack:api:update`. Everything on top is a patch, you can see how it's done for `chat_postMessage` in https://github.com/dblock/slack-ruby-client/tree/master/lib/slack/web/api/patches. 

Also will need a CHANGELOG entry, please.

I'll close this as is.

PR #105, opened 2016-08-26
Fix #104: thread leak in start_async with Celluloid.

### LINKED ISSUE #104: Thread leak in start_async
Every time the Slack client is reconnected, a new thread is created. Here is a code to reproduce the issue: https://gist.github.com/tikh/8d43a1a70dca55e89c0645cf7c9f6cc0. Thread count is going up for every reconnect. We are using celluloid-io gem here.

dblock:
Closes #104.

### DIFF
--- a/lib/slack/real_time/concurrency/celluloid.rb
+++ b/lib/slack/real_time/concurrency/celluloid.rb
@@ -38,6 +38,8 @@ module Slack
           rescue EOFError => e
             logger.debug("#{self.class}##{__method__}") { e }
             driver.emit(:close, WebSocket::Driver::CloseEvent.new(1001, 'server closed connection')) unless @closing
+          ensure
+            terminate
           end
 
           def close
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * [#103](https://github.com/slack-ruby/slack-ruby-client/pull/103): Added Danger, PR linting - [@dblock](https://github.com/dblock).
 * [#101](https://github.com/slack-ruby/slack-ruby-client/issues/101): Fix: protected method `close` called with EventMachine - [@dblock](https://github.com/dblock).
+* [#104](https://github.com/slack-ruby/slack-ruby-client/issues/104): Fix: thread leak in `start_async` with Celluloid - [@dblock](https://github.com/dblock).
 * Your contribution here.
 
 ### 0.7.6 (8/7/2016)


kandadaboggu:
Looks good. Let me try this out.

lib/slack/real_time/concurrency/celluloid.rb
```diff
@@ -38,6 +38,8 @@ def run_loop
           rescue EOFError => e
             logger.debug("#{self.class}##{__method__}") { e }
             driver.emit(:close, WebSocket::Driver::CloseEvent.new(1001, 'server closed connection')) unless @closing
+          ensure
+            terminate
```
kandadaboggu: Should this be done only for error exits? i.e.

```ruby
ensure
  if !$
    terminate
  end
end
```

dblock: I don't think so because in this case this class is the actor and there's literally nothing terminating it ever after even a "successful" run loop. Note that all run loops end with an EOF or some error, there's not even such a thing as clean termination. I could be wrong :)

kandadaboggu: I guess so. Since this is supposed to be a _forever_ loop may be we will never have successful exit.

I have tested the fix in my local setup. Looks like it is working.

BTW: Will there be a new release of the gem?

dblock: I'm pushing this fix to two production bots (https://www.playplay.io and https://market.playplay.io) and will cut a release before the end of the week if nothing implodes.

kandadaboggu: I think it might be good to check for the exception. I tried the code in the console. When I did an exit I got the following error:

```
Celluloid::DeadActorError: actor already terminated
       	/Users/harishs/.rvm/gems/ruby-2.3.0/gems/celluloid-0.17.3/lib/celluloid/proxy/actor.rb:34:in `terminate!'
       	/Users/harishs/.rvm/gems/ruby-2.3.0/gems/celluloid-0.17.3/lib/celluloid/proxy/cell.rb:64:in `terminate!'
       	/Users/harishs/.rvm/gems/ruby-2.3.0/gems/celluloid-0.17.3/lib/celluloid.rb:328:in `terminate'
       	/Users/harishs/.rvm/gems/ruby-2.3.0/gems/slack-ruby-client-0.7.4/lib/slack/real_time/concurrency/celluloid.rb:42:in `ensure in run_loop'
       	/Users/harishs/.rvm/gems/ruby-2.3.0/gems/slack-ruby-client-0.7.4/lib/slack/real_time/concurrency/celluloid.rb:42:in `run_loop'
```

dblock: Does something like `terminate unless terminated?` work there? How do you repro this?

dblock: Would you please PR ^^^? I would be happy with `terminate rescue Celluloid::DeadActorError nil` kind of thing too.

kandadaboggu: I will submit the PR with changes. I am testing a variation of the fix currently. 

PS: I reproduced the issue by running the code from ruby console. When you exit the console you will see the exception.

kandadaboggu: @dblock I have submitted the PR: https://github.com/slack-ruby/slack-ruby-client/pull/106

PR #106, opened 2016-08-26
Check for dead actors before termination

kandadaboggu:
Health of the actor is obtained using ` ::Celluloid::Actor.current.alive?`.

### DIFF
--- a/lib/slack/real_time/concurrency/celluloid.rb
+++ b/lib/slack/real_time/concurrency/celluloid.rb
@@ -39,7 +39,7 @@ module Slack
             logger.debug("#{self.class}##{__method__}") { e }
             driver.emit(:close, WebSocket::Driver::CloseEvent.new(1001, 'server closed connection')) unless @closing
           ensure
-            terminate
+            terminate if ::Celluloid::Actor.current.alive?
           end
 
           def close


lib/slack/real_time/concurrency/celluloid.rb
```diff
@@ -39,7 +39,7 @@ def run_loop
             logger.debug("#{self.class}##{__method__}") { e }
             driver.emit(:close, WebSocket::Driver::CloseEvent.new(1001, 'server closed connection')) unless @closing
           ensure
-            terminate
+            terminate if ::Celluloid::Actor.current.alive?
```
dblock: Is `::Celluloid::Actor.current` the same thing as `self` here? In which case `terminate unless alive?` would work better.

kandadaboggu: - When I use `alive?` and I keep getting 
```ruby
NoMethodError: undefined method `alive?' for #<Slack::RealTime::Concurrency::Celluloid::Socket:0x007fc8173f83d8>
```
For some strange reason `alive?` is resolved against `Slack::RealTime::Concurrency::Celluloid::Socket` instead of `Celluloid::Proxy::Cell(Slack::RealTime::Concurrency::Celluloid::Socket)`

- `Actor.current` refers to the current actor and it is thread safe.

https://github.com/celluloid/celluloid/blob/master/lib/celluloid/actor.rb#L17

dblock: Have you tried `proxy.alive?`

kandadaboggu: I have tried `alive?`, `behavior_proxy.alive?`, `proxy.alive?`. All of them fail.
E.g:
```
NameError: undefined local variable or method `proxy' for #<Slack::RealTime::Concurrency::Celluloid::Socket:0x007fb7ee7d0d68>
```

Interestingly, when I inspect the object it behaves differently,

```ruby
+2.3.0 :016 > slack_client.instance_variable_get("@socket").inspect
 => "#<Celluloid::Proxy::Cell(Slack::RealTime::Concurrency::Celluloid::Socket:0x3fc194e19f0c)..
```

My guess is when there is an un-handled exception proxy bindings are removed. But it still doesn't explain why `terminate` method is still accessible.

dblock:
What's being called here is https://github.com/celluloid/celluloid/blob/master/lib/celluloid.rb#L317 (found this via `p method(:terminate).source_location`), which does `Thread.current[:celluloid_actor].behavior_proxy.terminate!`. I made this `current_actor.terminate if current_actor.alive?` in https://github.com/slack-ruby/slack-ruby-client/commit/78da031cbe06361539b060fac97ea6a8ac18ee5a, LMK if I missed anything. 

Thanks for pointing this issue out and for your help!

kandadaboggu:
@dblock Tested locally. Your change works.

PR #112, opened 2016-10-06
add support for automated user invite

ohheyitsjeffrey:
There is an undocumented (well at least when I programmed it a few months back) method for automatically inviting users to slack via an api call with their email, which is very useful for support slack communities and corporate slack accounts.  

Let me know if there are any issues with this.  I finally just got around to committing this back to you, and thanks for this lovely ruby client.


dblock:
This code is auto-generated and would get overwritten over, see https://github.com/slack-ruby/slack-ruby-client/blob/master/CONTRIBUTING.md#update-slack-web-api.

ohheyitsjeffrey:
On it, thanks!

dblock:
I am going to close this since this is not going to be mergable.

Issue #110, opened 2016-10-04
RealTime Example behind Proxy


CanisHelix:
Based on advise from https://github.com/slack-ruby/slack-ruby-bot/issues/97 I have tested the hi_web example from behind my corporate firewall and it works without issue.

However the hi_real_time example does not, even when altered to include the proxy server in the configuration. Currently the only machines on the network with the proxy are headless, so I cannot test the normal Slack client either.

Am I configuring the proxy correctly here in this code?

**Log**:

```
I, [2016-10-04T03:53:18.471691 #6660]  INFO -- : post https://slack.com/api/rtm.start
D, [2016-10-04T03:53:18.471804 #6660] DEBUG -- request: Accept: "application/json; charset=utf-8"
User-Agent: "Slack Ruby Client/0.7.7"
Content-Type: "application/x-www-form-urlencoded"
I, [2016-10-04T03:53:18.875181 #6660]  INFO -- Status: 200
D, [2016-10-04T03:53:18.875356 #6660] DEBUG -- response: content-type: "application/json; charset=utf-8"
content-length: "3791"
connection: "close"
access-control-allow-origin: "*"
cache-control: "private, no-cache, no-store, must-revalidate"
content-security-policy: "referrer no-referrer;"
date: "Tue, 04 Oct 2016 03:53:18 GMT"
expires: "Mon, 26 Jul 1997 05:00:00 GMT"
pragma: "no-cache"
server: "Apache"
strict-transport-security: "max-age=31536000; includeSubDomains; preload"
vary: "Accept-Encoding"
x-accepted-oauth-scopes: "rtm:stream,client"
x-content-type-options: "nosniff"
x-oauth-scopes: "identify,read,post,client,apps"
x-slack-backend: "h"
x-slack-req-id: "e43ecc1d-be89-4f3a-8a80-cac8549171a9"
x-xss-protection: "0"
x-cache: "Miss from cloudfront"
via: "1.1 9f5fd09f16a7b31d703d6acac7f88026.cloudfront.net (CloudFront)"
x-amz-cf-id: "RAOXPlWc12pUzC7CdTWc7mOi5aIfTzxwAUyNkujUneU-QHJbqAiWiA=="
D, [2016-10-04T03:53:18.888880 #6660] DEBUG -- Slack::RealTime::Concurrency::Celluloid::Socket#connect!: WebSocket::Driver::Client
[...]

PR #121, opened 2016-11-09
Checking the current actor is running not just that it is alive

newdark:
The check for the alive should stay as it could be running but not
alive which is not likely but possible.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * [#116](https://github.com/slack-ruby/slack-ruby-client/pull/116): Use [slack-ruby/slack-api-ref](https://github.com/slack-ruby/slack-api-ref) as machine API reference - [@dblock](https://github.com/dblock).
 * [#116](https://github.com/slack-ruby/slack-ruby-client/pull/116): Added `users_setPhoto` and `users_deletePhoto` to Web API - [@dblock](https://github.com/dblock).
 * [#81](https://github.com/slack-ruby/slack-ruby-client/pull/81): Require faraday 0.9.0 or newer - [@leppert](https://github.com/leppert).
+* [#121](https://github.com/slack-ruby/slack-ruby-client/pull/121): Checking the current actor is running before calling terminate - [@newdark](https://github.com/newdark).
 * Your contribution here.
 
 ### 0.7.7 (8/29/2016)
--- a/lib/slack/real_time/concurrency/celluloid.rb
+++ b/lib/slack/real_time/concurrency/celluloid.rb
@@ -39,7 +39,7 @@ module Slack
             logger.debug("#{self.class}##{__method__}") { e }
             driver.emit(:close, WebSocket::Driver::CloseEvent.new(1001, 'server closed connection')) unless @closing
           ensure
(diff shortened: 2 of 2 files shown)


newdark:
@dblock Can you review this to support #110

dblock:
Is `alive?` and `running?` redundant? 

Is there a way we can write a test for this?

Needs a CHANGELOG entry either way.

newdark:
based on the code they are not even checking the same information.

https://github.com/celluloid/celluloid/blob/master/lib/celluloid/task.rb#L133

running? seams to check the status of the job. where alive is just checking if there is a actor. 

https://github.com/celluloid/celluloid/blob/master/lib/celluloid/mailbox.rb#L107

however if you don't check alive it will raise an error that says celluloid has to be alive. 

https://github.com/celluloid/celluloid/blob/master/lib/celluloid/proxy/actor.rb#L34

That check is not wrong but just extra insurance to make sure we don't terminate something that can't be terminated. Really it does not harm anything and if anything it is making things much more safe.

newdark:
Also tested this out on the server that was giving my the connection refused and it no longer gives me that error. 

Here is some code of a terminate that does a better job of checking if can be terminated. 

https://github.com/celluloid/celluloid/blob/master/lib/celluloid/task.rb#L108

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * [#116](https://github.com/slack-ruby/slack-ruby-client/pull/116): Use [slack-ruby/slack-api-ref](https://github.com/slack-ruby/slack-api-ref) as machine API reference - [@dblock](https://github.com/dblock).
 * [#116](https://github.com/slack-ruby/slack-ruby-client/pull/116): Added `users_setPhoto` and `users_deletePhoto` to Web API - [@dblock](https://github.com/dblock).
 * [#81](https://github.com/slack-ruby/slack-ruby-client/pull/81): Require faraday 0.9.0 or newer - [@leppert](https://github.com/leppert).
+* [#121](https://github.com/slack-ruby/slack-ruby-client/pull/121): Checking the current actor is running before calling terminate
```
dblock: Can you please fix the format for this? Needs your name at the end with a period and all ...

dblock:
I'd like to merge this, it needs to be rebased. Also any ideas about a test @newdark? Maybe a connection against a server that doesn't exist?

newdark:
Yes total forgot about this work. Thanks for the reminder. I think that moving work I will have to test into out. I also noticed that there was no tests for this file is there a style guide to follow for writing tests?

dblock:
Sorry no style guide. There should be tests for everything obviously.

newdark:
@dblock

Ok I will make sure to get this all corrected. Been a bit busy but once I get some free time I can have it done.

dblock:
Bump @newdark

newdark:
Sorry I have been a bit over worked with my job @dblock I have not had time to come back to work on this :(

dblock:
I failed to write a good test for this, so I am going to merge via https://github.com/slack-ruby/slack-ruby-client/pull/137.

PR #123, opened 2016-11-14
Stop a duplicate definition warning

michaelherold:
When running ruby with `RUBYOPT="-w"` or with RSpec running with
warnings turned on, these errors were being thrown:

```
slack-ruby-client/lib/slack/real_time/concurrency/celluloid.rb:45: warning: method redefined; discarding old close
ruby-2.3.1/lib/ruby/2.3.0/forwardable.rb:187: warning: previous definitionof close was here
slack-ruby-client/lib/slack/real_time/concurrency/celluloid.rb:61: warning: method redefined; discarding old write
ruby-2.3.1/lib/ruby/2.3.0/forwardable.rb:187: warning: previous definitionof write was here
```

This change removes the delegators that were present, since the wrappers
call them directly anyway.


dblock:
Thanks @michaelherold. Wonder whether we should run rspec with `-w` in tests and fail on warnings?

michaelherold:
I typically try to run with this configuration:

```ruby
RSpec.configure do |config|
  config.warnings = true
end
```

That makes a lot of things pop in the library (mostly uninitialized instance variables), which is helpful from a code quality standpoint.

However, in this case, there are a lot of warnings from the `picky` gem as well, so it will definitely muddle the RSpec output, at least until we go fix those issues in the related library.

Do you think that would be worth it? It at least shows where the squeaky wheels are.

dblock:
Probably not, thanks @michaelherold.

Issue #134, opened 2017-02-10
RTM connection fails in large org but works in small org


geekdave:
Hello,

I'm attempting to use a bot to connect Slack with Pagerduty: https://github.com/stripe-contrib/pagerbot

* I set up the bot with a Slack bot access token.  
* I verified that the token works by using "curl" from the command line and was able to see the message getting posted in Slack.
* I tested the bot on a large org (5000+ people) and it consistently fails to connect.
* I tested the bot on a small org (15 people) (using a different token of course), and it worked fine. 

I've worked with the developer of the bot @macobo, and he suggested I file an issue here.  

[I've created a gist with the error logs from heroku](https://gist.github.com/geekdave/71c99db97156e21d83e4d83deffb6b99).  The first log shows the errors when running this in the large org.  The second log shows the successful connection with a smaller org:

I added some comments at the bottom noting some differences, notably:

1) On large org I see this in the response, but not in the small org:

```
transfer-encoding: "chunked" 
```

Could there be an issue with the large number of members in the org or the number of channels causing something to stall?  Is there a huge payload coming back from the RTM API due to my large org?

2) There is a delay of 100+ seconds in the large org log, and the error:

```
Socket URL has expired
```

Thanks for looking!

geekdave:
UPDATE: I pinged Slack's API help team about this and got the following response:

```
My guess is the bot is timing out on the initial call to rtm.start, which is the method that's called to initialize an RTM connection — for large teams this can take a long time. Thinking aloud, it could also be a memory allocation issue (depending on your server config).

Any chance you can increase the timeout for the initial API call ?

We're working very (very) hard on improvements in this area, as you can imagine this is a performance problem across-the-board for very large teams.
```

/cc @macobo

dblock:
Indeed, there's a very large payload coming back from the initial `rtm.start` with all the org information. Tweak the call to `rtm.start` with `simple_latest: true` and `no_unreads`, that reduces the size by a bit.

Then disable any client-side parsing of that data with `Slack::RealTime::Client.new(store_class: nil)`, see if that helps at all.

dblock:
There's a feature request on supplying a custom Faraday connection in https://github.com/slack-ruby/slack-ruby-client/issues/72. In the meantime you can probably test other connection options with some editing of https://github.com/slack-ruby/slack-ruby-client/blob/master/lib/slack/web/faraday/connection.rb or monkey-patching.

macobo:
[...]

PR #135, opened 2017-03-08
Added timeout and open_timeout options to Web API.

dblock:
Related to #134.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ### 0.8.0 (Next)
 
+* [#134](https://github.com/slack-ruby/slack-ruby-client/issues/134): Added `timeout` and `open_timeout` options to Web API - [@dblock](https://github.com/dblock).
 * Your contribution here.
 
 ### 0.7.9 (2/9/2017)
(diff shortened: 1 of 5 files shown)


macobo:
Are there any downsides to also increasing the default?

dblock:
Faraday doesn't default the value to anything, letting the underlying Net::HTTP library handle this. I am a little wary setting it to something, but maybe I can be convinced otherwise.

dblock:
Check out my note in README though.

macobo:
:+1: This is definitely an improvement, thanks for fixing. 

One upside of a larger default is that if this extension is ever used in say an open-source chatbot ([pagerbot](https://github.com/stripe-contrib/pagerbot/)) having a large problem would avoid per-bot hacks and troubles for some of the larger users who try to use the bot and fail to log in.

It might be worthwhile to merge this, then reach out to some of the people having issues and figuring out if a larger but still sane baseline setting as a default would make sense.

dblock:
I agree, we can save a lot of people debugging. What value are you using @macobo? Also are you changing only `timeout` or `open_timeout` as well?

dblock:
Oh according to https://ruby-doc.org/stdlib-2.4.0/libdoc/net/http/rdoc/Net/HTTP.html both of these are 60 (!) seconds.

macobo:
Sadly I don't have access to a large organization to test with, sorry!

PR #137, opened 2017-03-08
Fix: check that the current Celluloid actor is running before calling terminate.

dblock:


newdark:
@dblock  The link of code that it is calling is here. https://github.com/celluloid/celluloid/blob/master/lib/celluloid/task.rb#L139

dblock:
@newdark Check out my spec in https://github.com/slack-ruby/slack-ruby-client/pull/147. Try putting it back.

PR #151, opened 2017-05-22
CI against Ruby 2.4.1

syguer:
Ruby 2.4.1 released
Ref https://www.ruby-lang.org/en/news/2017/03/22/ruby-2-4-1-released/


syguer:
I think no need to update CHANGELOG.md. does need?

dblock:
I'll take it, thanks!

PR #154, opened 2017-07-25
Raise a dedicated error class with 429 responses

### LINKED ISSUE #153: Check Retry-After header
I accidentally got rate limited when I enqueued a number of jobs to hit the Slack API and I'm trying to comply with the rate limit, however I don't see any existing means of accessing the `Retry-After` header [mentioned in the docs](https://api.slack.com/docs/rate-limits).  Am I missing something or should I open a PR to better expose this value in the event of a 429 response code?

greggroth:
Ensure the error message includes the value of the "Retry-After" header
so the client knows when it's safe to retry requests.

Closes #153


spec/slack/web/api/endpoints/custom_specs/auth_spec.rb
```diff
@@ -13,8 +13,13 @@
     end
   end
   context '429 error', vcr: { cassette_name: 'web/429_error' } do
-    it 'fails with an exception' do
-      expect { client.auth_test }.to raise_error Faraday::ClientError
+    it 'fails with an specific exception' do
+      begin
+        client.auth_test
+      rescue Slack::Web::Api::TooManyRequestsError => e
```
greggroth: Note that this new error class does _not_ inherit from `Faraday::ClientError`, so a client is expecting to rescue that error in the case of a 429 response code, this PR will break that behavior.

lib/slack-ruby-client.rb
```diff
@@ -19,6 +19,7 @@
 end
 require_relative 'slack/web/config'
 require_relative 'slack/web/api/error'
+require_relative 'slack/web/api/too_many_requests_error'
```
dblock: Lets move it to an `errors` folder?

lib/slack/web/api/too_many_requests_error.rb
```diff
@@ -0,0 +1,21 @@
+module Slack
+  module Web
+    module Api
+      class TooManyRequestsError < ::Faraday::Error
```
dblock: `module Errors`

dblock:
This is great, just move it into a folder structure for future errors and I'll merge.

lib/slack/web/api/too_many_requests_error.rb
greggroth: 👍

greggroth: It may be confusing that this error is namespaced under `Slack::Web::Api::Errors` while the existing `Slack::Web::Api::Error` class is not namespaced (and is very close to `Slack::Web::Api::Errors`).  Do you still want the change?

dblock: I think so. It's not uncommon. We can also do a `Slack::Web::Api::Errors::Base` and alias the old one for backward compatibility if you want to, your call.

dblock:
Looks great, merging.

Issue #158, opened 2017-08-16
update for latest slack-api-ref


alexagranov:
About to open a PR with new methods, updates to patches, and tests.

PR #159, opened 2017-08-17
Issue #158 - updating to latest slack-api-ref

alexagranov:
I don't believe there were any breaking changes introduced - but of course, we'll need to test.

Note there are 2 commits so that I could generate a proper standalone diff for the `postEphemeral` patch.

Along with CHANGELOG comments, added specs for `postEphemeral` that reflect the required options.


dblock:
Looks great, thanks for the update @alexagranov. Adding you here as well if you want to make a release.

alexagranov:
@dblock Thanks, you paved the way ;-)  Let me do some testing with my servers, then a couple days to catch the eclipse 🔭  If all goes well, I'll make a release next week.

PR #162, opened 2017-09-06
Gracefully close websocket on Errno::EPIPE

johanoskarsson:
I kept running into the following error:
E, [2017-09-05T14:23:29.153409 #10948] ERROR -- : Actor crashed!
Errno::EPIPE: Broken pipe
        /usr/share/rvm/rubies/ruby-2.3.4/lib/ruby/2.3.0/openssl/buffering.rb:379:in `syswrite_nonblock'

This would crash my bot entirely. By catching Errno::EPIPE and gracefully closing (and later reconnecting) the error has gone away. I have not done any deeper digging into this issue, but in case others run into it this worked for me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### 0.9.2 (Next)
 
 * [#161](https://github.com/slack-ruby/slack-ruby-client/pull/161): Added support for cursor pagination - [@dblock](https://github.com/dblock).
+* [#162](https://github.com/slack-ruby/slack-ruby-client/pull/162): Gracefully close websocket on errno::epipe - [@johanoskarsson](https://github.com/johanoskarsson).
 * Your contribution here.
 
 ### 0.9.1 (8/24/2017)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ### 0.9.2 (Next)
 
 * [#161](https://github.com/slack-ruby/slack-ruby-client/pull/161): Added support for cursor pagination - [@dblock](https://github.com/dblock).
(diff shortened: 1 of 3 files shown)


dblock:
This looks good. Needs a CHANGELOG entry and a test ala https://github.com/slack-ruby/slack-ruby-client/blob/9ed4d21a5fced74734eb43cf7371509fbf27d388/spec/slack/real_time/concurrency/celluloid_spec.rb, please.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ### 0.9.2 (Next)
 
 * [#161](https://github.com/slack-ruby/slack-ruby-client/pull/161): Added support for cursor pagination - [@dblock](https://github.com/dblock).
+* [#162](https://github.com/slack-ruby/slack-ruby-client/pull/162): Gracefully close websocket on errno::epipe - [@johanoskarsson](https://github.com/johanoskarsson).
```
dblock: Can I be difficult and nitpick? The correct class is `Errno::EPIPE`, it's case sensitive, so quote it as code and correct upper/lowercase please.

spec/slack/real_time/concurrency/celluloid_spec.rb
```diff
@@ -43,6 +43,22 @@ def read
             socket.run_loop
           end
         end
+
+        describe '#run_loop with epipe' do
+          let(:test_socket) do
```
dblock: Lets reorganize this a bit? There should be a default test socket for a bunch of tests. Then the `fail` scenarios should be separate with something like 

```ruby
[Errno::EPIPE,  EOFError].each do |ex|
 context "when #{ex} is raised" do
 ...
 end
end
```

dblock:
Merged via https://github.com/slack-ruby/slack-ruby-client/commit/1a68e2b7fc06c78444cd5b25f905a0d5694d6a95.

PR #163, opened 2017-09-12
Fix logging errors in real-time component

rfwatson:
Avoids `log writing failed. "\x81 from ASCII-8BIT to UTF8"`

### DIFF
--- a/lib/slack/real_time/concurrency/celluloid.rb
+++ b/lib/slack/real_time/concurrency/celluloid.rb
@@ -53,12 +53,12 @@ module Slack
           end
 
           def handle_read(buffer)
-            logger.debug("#{self.class}##{__method__}") { buffer }
+            logger.debug("#{self.class}##{__method__}") { buffer.force_encoding('UTF-8') }
             driver.parse buffer
           end
 
           def write(data)
-            logger.debug("#{self.class}##{__method__}") { data }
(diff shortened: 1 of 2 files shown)


dblock:
Is it documented that this data comes in as UTF-8? It would be nice to have a test for this, please.

lib/slack/real_time/concurrency/celluloid.rb
```diff
@@ -53,12 +53,12 @@ def read
           end
 
           def handle_read(buffer)
-            logger.debug("#{self.class}##{__method__}") { buffer }
+            logger.debug("#{self.class}##{__method__}") { buffer.force_encoding('UTF-8') }
```
mikz: I think it is better to use `scrub` (if it is available on all supported ruby versions).

Or possibly even `begin` & `rescue` so the logging can't crash reading/writing to a socket.

rfwatson:
Looked into this issue a bit further:

* It appears to affect both celluloid and eventmachine components
* Causes logging of most incoming buffers to fail
* Caused by 2-4 bytes of non UTF8-compatible padding which appears to be prepended to incoming buffers  (I'm not clear what this padding actually is??)

However, when trying to create a minimal example outside a Rails env, the buffer seems to be logged successfully, binary padding and all. Given that it therefore seems likely to be an issue with ActiveSupport logger config, I'm closing this PR.

PR #167, opened 2017-09-18
respect rate limits during pagination by sleeping

### LINKED ISSUE #166: rate limiting with pagination
For large teams paginated calls to `users.list` and other methods will inevitably lead to rate limiting (429 responses).

The current implementation of pagination doesn't offer any way to handle this gracefully.

jmanian:
resolves #166 

Here's an initial attempt at respecting rate limits during pagination. No tests here yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ### 0.9.2 (Next)
 
-* [#167](https://github.com/slack-ruby/slack-ruby-client/pull/167): Respect rate limits during pagination by sleeping. Also add optional `pause` parameter in order to proactively sleep between each paginated request - [@jmanian](https://github.com/jmanian).
+* [#167](https://github.com/slack-ruby/slack-ruby-client/pull/167): Added support for pausing between paginated requests that can cause Slack rate limiting - [@jmanian](https://github.com/jmanian).
 * [#163](https://github.com/slack-ruby/slack-ruby-client/pull/164): Use `OpenSSL::X509::DEFAULT_CERT_DIR` and `OpenSSL::X509::DEFAULT_CERT_FILE` for default ca_cert and ca_file - [@leifcr](https://github.com/leifcr).
 * [#161](https://github.com/slack-ruby/slack-ruby-client/pull/161): Added support for cursor pagination - [@dblock](https://github.com/dblock).
 * Your contribution here.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ### 0.9.2 (Next)
 
+* [#167](https://github.com/slack-ruby/slack-ruby-client/pull/167): Respect rate limits during pagination by sleeping. Also add optional `pause` parameter in order to proactively sleep between each paginated request - [@jmanian](https://github.com/jmanian).
 * [#163](https://github.com/slack-ruby/slack-ruby-client/pull/164): Use `OpenSSL::X509::DEFAULT_CERT_DIR` and `OpenSSL::X509::DEFAULT_CERT_FILE` for default ca_cert and ca_file - [@leifcr](https://github.com/leifcr).
 * [#161](https://github.com/slack-ruby/slack-ruby-client/pull/161): Added support for cursor pagination - [@dblock](https://github.com/dblock).
 * Your contribution here.
--- a/lib/slack/web/pagination/cursor.rb
+++ b/lib/slack/web/pagination/cursor.rb
@@ -19,11 +19,15 @@ module Slack
             next_cursor = nil
             loop do
               query = { limit: client.default_page_size }.merge(params).merge(cursor: next_cursor)
-              response = client.send(verb, query)
-              yield response
-              break unless response.response_metadata
-              next_cursor = response.response_metadata.next_cursor
-              break if next_cursor.blank?
+              begin
+                response = client.send(verb, query)
+                yield response
+                break unless response.response_metadata
+                next_cursor = response.response_metadata.next_cursor
+                break if next_cursor.blank?
+              rescue Slack::Web::Api::Errors::TooManyRequestsError => e
+                sleep(e.retry_after.seconds)
+              end
             end
           end
         end
--- a/lib/slack/web/pagination/cursor.rb
+++ b/lib/slack/web/pagination/cursor.rb
@@ -7,11 +7,13 @@ module Slack
 
           attr_reader :client
           attr_reader :verb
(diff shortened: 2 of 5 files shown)


jmanian:
That's a good idea about hinting a sleep value, though we may find it hard to tune to avoid the errors. The reason I say this is I asked some Slack folks about their rate limiting a while back and they had this to say:

> Our rate limiting is pretty brute. At the same time, we don't consider getting rate limited a ding or demerit. Some devs might just want to burn through as fast as possible until hitting a rate limit, then pausing activity before resuming, theoretically, after following the Retry-after. ([source](https://dev4slack.slack.com/archives/C0K8617D5/p1502148856419895?thread_ts=1502148468.326433&cid=C0K8617D5))

> The specific rate limits for the web API aren't actually published and can vary somewhat method to method, with some burst behavior allowed here and there. By virtue of being unpublished, the best recommendation is to make requests until you hit a limit, then backoff until the time the `retry-after` indicates. ([source](https://dev4slack.slack.com/archives/C0K8617D5/p1502206135557519?thread_ts=1502148468.326433&cid=C0K8617D5))

Given this context, I think the best solution is to allow the individual dev to pass in their own default sleep time, so that they can tune to their use case, but not have any default sleep behavior beyond the rate limiting.

dblock:
I agree, lets default the hint to zero. 

I wonder whether retry behavior could be optionally built into everything? As a client I just don't want to deal with it :) But that can be a later PR/feature - feel free to open an issue for that too.

jmanian:
I think this is what you had in mind for hinting a sleep value. I went with `pause` for the parameter so as to not overload `sleep` with an attribute on top of the method, but I'm happy to change this.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ### 0.9.2 (Next)
 
+* [#167](https://github.com/slack-ruby/slack-ruby-client/pull/167): Respect rate limits during pagination by sleeping. Also add optional `pause` parameter in order to proactively sleep between each paginated request - [@jmanian](https://github.com/jmanian).
```
dblock: Maybe simply "Added support for pausing between paginated requests that cause Slack rate limiting"?

lib/slack/web/pagination/cursor.rb
```diff
@@ -7,23 +7,31 @@ class Cursor
 
           attr_reader :client
           attr_reader :verb
+          attr_reader :pause
```
dblock: I feel a bit uncomfortable calling this `pause` because it's unclear whether that's a verb or a noun, maybe `sleep_interval`? Don't feel strongly about this one though.

lib/slack/web/pagination/cursor.rb
```diff
@@ -7,23 +7,31 @@ class Cursor
...
           def each
             next_cursor = nil
             loop do
               query = { limit: client.default_page_size }.merge(params).merge(cursor: next_cursor)
-              response = client.send(verb, query)
+              begin
+                response = client.send(verb, query)
+              rescue Slack::Web::Api::Errors::TooManyRequestsError => e
+                sleep(e.retry_after.seconds)
+                next
```
dblock: I also think we should have a max retry count with some default set fairly high. If slack blacklists you somehow you'll have code that never ends and it will be impossible to debug.

I also think there should be a log line here in DEBUG mode.

jmanian: Do you think the count should reset after a successful request?

dblock: I think so.

lib/slack/web/pagination/cursor.rb
```diff
@@ -7,23 +7,38 @@ class Cursor
...
           attr_reader :client
           attr_reader :verb
+          attr_reader :sleep_interval
+          attr_reader :max_retries
           attr_reader :params
 
           def initialize(client, verb, params = {})
             @client = client
             @verb = verb
+            @sleep_interval = params.delete(:sleep_interval)
```
dblock: This modifies incoming `params`, ie. has a side-effect, you don't want to do that, so just `params[:sleep_interval]`.

lib/slack/web/pagination/cursor.rb
```diff
@@ -7,23 +7,38 @@ class Cursor
...
+              begin
+                response = client.send(verb, query)
+              rescue Slack::Web::Api::Errors::TooManyRequestsError => e
+                raise e if retry_count >= max_retries
+                client.logger.debug("#{self.class}##{__method__}") { e.to_s }
+                retry_count += 1
+                sleep(e.retry_after.seconds)
+                next
+              end
               yield response
```
dblock: You can move this inside the rescue block and avoid `next`, right?

dblock: Also did I tell you that I was a nitpicky code reviewer? Hang on tight :)

jmanian: I understand, I'm the same way 😎.

Which part are you suggesting go in the rescue block? I started out with everything inside the begin block (see first commit b75bfcb) which avoids the `next`. Is this what you mean?

I changed it (8542d00) because I thought it was bad practice to have more lines than necessary inside the `begin` block — even if just for clarity (so that it's clear which line is expected to raise the error). But in this case it's obvious where the error is coming from, so I guess it's fine either way.

dblock: Exactly.

dblock:
Merged. Thanks!

lib/slack/web/pagination/cursor.rb
dblock: I actually now think yours is better. Someone can do something inside the yield that causes that exception and we're going to have bad side effects and catching an error we shouldn't.

jmanian:
I was debating just leaving those parameters in there, because I think all the API methods ignore those extraneous parameters. But it felt more right to remove them.

dblock:
I opened #168 - interested in giving that one a try?

kyanh-huynh:
@dblock  Is this problem related to login/authentication stage? I noticed that when I deployed the bot (`ruby-slack-bot`), the bot can be connected to Slack Api. (well, may laptop network is slow)

However, on AWS ec2 where the network is extremely fast, the first 2 requests to Slack happens in within 1 second (I can see this when I turn on DEBUG mode), and then Slack simply blocks the future 3rd request. The bot will never be able to authenticate to the service.

dblock:
@kyanh-huynh It would mitigate that as well.

Issue #170, opened 2017-09-19
Release 0.10.0


dblock:

jmanian:
In addition to the usual steps I should also change the version in [lib/slack/version.rb](https://github.com/slack-ruby/slack-ruby-client/blob/master/lib/slack/version.rb) (and in the changelog) to 0.10.0 *before* releasing, is that right?

dblock:
Correct @jmanian

jmanian:
I'll wait for #171 to merge then I'll rip it.

dblock:
All merged.

jmanian:
Ah shoot, I just realized I used yesterday's date in the changelog for the release (because I initially made the commit yesterday, then rebased after those PRs). This means the date in the changelog doesn't match the date on rubygems.

I think there are 4 options:
1. Leave it be
1. Update the date in the changelog in a new commit (which means [the date in v0.10.0](https://github.com/slack-ruby/slack-ruby-client/blob/v0.10.0/CHANGELOG.md) will always be wrong).
1. Amend the commit (which means the v0.10.0 tag won't identically match the published gem — they'll have different SHAs).
1. Amend the commit *and* republish the gem (the nuclear option).

dblock:
Leave it.

PR #172, opened 2017-09-19
Use rtm.start when store is set to any subclass of Slack::RealTime::Stores::Store

kstole:
Currently, the default `store_class` is `Slack::RealTime::Store`, but `rtm_start_method` checks for `Slack::RealTime::Stores::Store` when selecting `:rtm_start`. This change allows subclasses to also use `:rtm_start`.


dblock:
This needs a test and a CHANGELOG entry please. Might make it into 0.10.0 if you hurry :) Re: #170

kstole:
Updated tests. The client_spec was a bit out of order and I didn't overhaul it fully but hopefully an improvement overall 👍

dblock:
Very good, much thanks.

PR #176, opened 2017-10-29
Remove json gem from dependencies

seuros:


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ### 0.10.1 (next)
 
 * Your contribution here.
+* [#176](https://github.com/slack-ruby/slack-ruby-client/issues/176): Drop dependency on json gem. - [@seuros](https://github.com/seuros).
```
dblock: You have an extra period after `gem.`, so Danger complains.

dblock:
Did this come up as an issue somewhere, just curious. Obviously we don't need it if everything turns green.

seuros:
@dblock. Thank you for the review.
The json library is part of ruby now. 
I did notice the dependency when one of my build failed when the json library failed to fetch from rubygems.

PR #177, opened 2017-11-06
fix api ref repo

manuelmeurer:
It seems this was changed from https://github.com/dblock/slack-api-ref to https://github.com/slack-ruby/slack-api-ref a while ago...


dblock:
See the Danger note above please, otherwise looks good.

manuelmeurer:
I updated the changelog after seeing the Danger note and force-pushed. Should it not check the updated commit again?

dblock:
Thanks!

PR #179, opened 2017-11-20
update to latest slack api ref

jmanian:
For some reason the methods have all changed order. Any idea why?
As a result I had to redo one of the patches, `chat.3`, because the line numbers were no longer correct.

There's one new endpoint buried in here amongst the noise of reordering, `migration.exchange`.


jmanian:
I'm looking at the diff closer and it looks like everything used to be alphabetical, but not anymore.

jmanian:
I'll give that a shot.

dblock:
Replaced with https://github.com/slack-ruby/slack-ruby-client/pull/181.

PR #181, opened 2017-11-25
Update slack api ref

dblock:
Replaces https://github.com/slack-ruby/slack-ruby-client/pull/179, for now no undocumented method support.


dblock:
The build error is some transient problem with `echo.websocket.org`, merging.

PR #182, opened 2017-11-25
Add undocumented methods

dblock:


dblock:
The build failure is transient, opened #183.

PR #185, opened 2017-11-25
Updated method template to include a warning for undocumented endpoints

aviflombaum:


aviflombaum:
I updated the method template to include a warning to the logger for undocumented methods, wasn't quite sure how to test this as everything is autogenerated, but I was able to see the warning log happen when using this client. I'd imagine the test might be on something like the method_spec template adding a new test that `logger` should receive a call to `warn` if the method is undocumented but the tests seem to be higher level. Anyway, point me in the right direction, I'll do my best to complete this addition.

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ### 0.11.1 (Next)
 
-* Your contribution here.
```
dblock: Put this back please.

lib/slack/web/api/templates/method.erb
```diff
@@ -44,6 +44,9 @@ module Slack
 <% if data['args']['user'] %>
             options = options.merge(user: users_id(options)['user']['id']) if options[:user]
 <% end %>
+<% if data['undocumented'] %>
+            logger.warn('<%= group %>.<%= name %> is an undocumented API endpoint.')
```
dblock: Maybe "The <%= group %>.<%= name %> method is undocumented."

aviflombaum:
I'm mostly confused as to where to code the tests - should I alter the `method_spec.erb` template to generate tests that the undocumented call `warn` on `logger`? Or is there a higher level place to code this test? I'll also make an attempt at some approach tomorrow.

I'll look until a silence option on the `Slack::Logger` so that it can be configured to silence warnings. 

Will clean it up and squash commits too.

lib/slack/web/api/templates/method.erb
aviflombaum: 👍

dblock:
You could just add specs into https://github.com/slack-ruby/slack-ruby-client/blob/master/spec/slack/web/client_spec.rb or add an undocumented_spec.rb picking one of the undocumented methods and ensuring `logger.warn` is called, etc.

dblock:
Closing in favor of https://github.com/slack-ruby/slack-ruby-client/pull/197. @aviflombaum Take a look, is that what you had in mind? Make comments on that issue.

PR #187, opened 2017-11-29
Concatenate error message when multiple errors present

chrislopresto:
In the else clause of the `Faraday::Response::RaiseError#on_complete ` method, we expect that the body object will have an `'error'` key. However, sometimes a response will have an `'errors'` key (plural) that contains a collection of error responses. 

I encountered this when making calls to `client.conversations_invite` with multiple users:

```sh
> client.conversations_invite users: [paul.slack_user_id, ringo.slack_user_id].join(','), channel: '#project-beatles'
----------------
Printing body.inspect to console
#<Slack::Messages::Message errors=#<Hashie::Array [#<Slack::Messages::Message error="already_in_channel" ok=false user="xxxxxxxxx">, #<Slack::Messages::Message error="already_in_channel" ok=false user="xxxxxxxxx">]> ok=false>
----------------
Slack::Web::Api::Errors::SlackError: Slack::Web::Api::Errors::SlackError
from /Users/chrislopresto/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/slack-ruby-client-0.11.0/lib/slack/web/faraday/response/raise_error.rb:18:in `on_complete'
```

As a result, we don't pass up any helpful error message.

I'm not sure what the desired behavior should be when there are multiple errors. The tweak in this PR just concatenates an error message. So the results of the same call would now be:

```sh
> client.conversations_invite users: [paul.slack_user_id, george.slack_user_id].join(','), channel: '#project-beatles'
Slack::Web::Api::Errors::SlackError: already_in_channel,already_in_channel
from /Users/chrislopresto/.rbenv/versions/2.4.2/lib/ruby/gems/2.4.0/gems/slack-ruby-client-0.11.0/lib/slack/web/faraday/response/raise_error.rb:13:in `on_complete'
```

Let me know if this makes sense, if you want prefer another approach, if we should add a spec, etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ### 0.11.1 (Next)
 
+* [#187](https://github.com/slack-ruby/slack-ruby-client/pull/187): Concatenate error message when multiple errors present - [@chrislopresto](https://github.com/chrislopresto).
 * Your contribution here.
 
 ### 0.11.0 (11/25/2017)
--- a/lib/slack/web/faraday/response/raise_error.rb
+++ b/lib/slack/web/faraday/response/raise_error.rb
@@ -9,7 +9,8 @@ module Slack
             elsif (body = env.body) && body['ok']
               nil
             else
-              raise Slack::Web::Api::Errors::SlackError.new(body['error'], env.response)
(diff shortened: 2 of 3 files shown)


dblock:
This is a reasonable change, but it needs tests, please.

Issue #32, opened 2015-11-23
NoMethodError in replaced raise_error method for Faraday::Response::RaiseError


contentfree:
Stacktrace:

```shell
/usr/local/rvm/gems/ruby-2.1.4/gems/slack-ruby-client-0.4.0/lib/slack/web/faraday/response/raise_error.rb:8:in `on_complete': undefined method `[]' for nil:NilClass (NoMethodError)
	from /usr/local/rvm/gems/ruby-2.1.4/gems/faraday-0.9.2/lib/faraday/response.rb:9:in `block in call'
	from /usr/local/rvm/gems/ruby-2.1.4/gems/faraday-0.9.2/lib/faraday/response.rb:57:in `on_complete'
	from /usr/local/rvm/gems/ruby-2.1.4/gems/faraday-0.9.2/lib/faraday/response.rb:8:in `call'
	from /usr/local/rvm/gems/ruby-2.1.4/gems/faraday-0.9.2/lib/faraday/response.rb:8:in `call'
	from /usr/local/rvm/gems/ruby-2.1.4/gems/faraday-0.9.2/lib/faraday/request/url_encoded.rb:15:in `call'
	from /usr/local/rvm/gems/ruby-2.1.4/gems/faraday-0.9.2/lib/faraday/request/multipart.rb:14:in `call'
	from /usr/local/rvm/gems/ruby-2.1.4/gems/faraday-0.9.2/lib/faraday/rack_builder.rb:139:in `build_response'
	from /usr/local/rvm/gems/ruby-2.1.4/gems/faraday-0.9.2/lib/faraday/connection.rb:377:in `run_request'
	from /usr/local/rvm/gems/ruby-2.1.4/gems/faraday-0.9.2/lib/faraday/connection.rb:177:in `post'
	from /usr/local/rvm/gems/ruby-2.1.4/gems/slack-ruby-client-0.4.0/lib/slack/web/faraday/request.rb:25:in `request'
	from /usr/local/rvm/gems/ruby-2.1.4/gems/slack-ruby-client-0.4.0/lib/slack/web/faraday/request.rb:10:in `post'
	from /usr/local/rvm/gems/ruby-2.1.4/gems/slack-ruby-client-0.4.0/lib/slack/web/api/endpoints/channels.rb:131:in `channels_list'
```

A guard needs to be added to handle the case when `body` isn't present

dblock:
[...]

PR #188, opened 2017-11-30
Fix NoMethodError when Slack is unavailable

sonicdoe:
When Slack is experiencing connectivity issues and returns a `503 Service Unavailable`, slack-ruby-client should throw a `Faraday::ClientError`. However, due to https://github.com/slack-ruby/slack-ruby-client/blob/ffe7ec87ef74d99d2f162b05c976dc604d3aeb99/lib/slack/web/faraday/response/raise_error.rb#L12 a ``NoMethodError: undefined method `[]' for nil:NilClass`` is raised (since `body` is `nil`). The same error was also reported in https://github.com/slack-ruby/slack-ruby-client/issues/32.

I haven’t looked into an actual fix yet but [I figured](https://twitter.com/sindresorhus/status/579306280495357953) submitting a failing test would help 🙂

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### 0.11.1 (Next)
 
 * [#187](https://github.com/slack-ruby/slack-ruby-client/pull/187): Concatenate error message when multiple errors present - [@chrislopresto](https://github.com/chrislopresto).
+* [#188](https://github.com/slack-ruby/slack-ruby-client/pull/188): Fixed `NoMethodError` when Slack is unavailable - [@sonicdoe](https://github.com/sonicdoe).
 * Your contribution here.
 
 ### 0.11.0 (11/25/2017)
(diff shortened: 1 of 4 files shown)


dblock:
Good call! We generally need to handle 50x error codes differently from 40x error codes that contain a JSON body.

sonicdoe:
By the way, that’s also on the roadmap for Faraday v1.0 (see https://github.com/lostisland/faraday/pull/735).

sonicdoe:
I’ve now fixed the `NoMethodError` by changing the logic in [`raise_error.rb`](https://github.com/slack-ruby/slack-ruby-client/blob/ffe7ec87ef74d99d2f162b05c976dc604d3aeb99/lib/slack/web/faraday/response/raise_error.rb). It now raises a `SlackError` only if `body['ok']` is `false`. If `body` is `nil`, the response will fall through to `::Faraday::Response::RaiseError` which will then raise a `Faraday::ClientError`.

dblock:
Perfect, care to rebase this please and I'll merge.

dblock:
Thanks, sorry didn't get a notification on rebase.

sonicdoe:
> Maybe describe error handling in README if you think it can be helpful to others, too.

I’ve now opened #190 which documents all of slack-ruby-client’s error handling.

PR #206, opened 2018-04-29
Fix 100% CPU usage in async examples

felixbuenemann:
This switches from the endless loop to accumulating and joining the threads.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### 0.11.2 (Next)
 
 * Your contribution here.
+* [#206](https://github.com/slack-ruby/slack-ruby-client/pull/206): Fix 100% cpu usage in async examples - [@felixbuenemann](https://github.com/felixbuenemann).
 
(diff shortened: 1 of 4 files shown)


felixbuenemann:
Another approach would be to insert something like `sleep 0.1` into the loop, but I think this one is both more correct and more instructive, since it shows that `start_async` returns a thread.

I only tested this code with the celluloid implementation when fixing will/slacktyping#1, but the EventMachine implementation also returns a thread so it should be fine.

felixbuenemann:
Added suggested changelog entry.

dblock:
Thanks! Looks like build failed for something unrelated, a library upgrade, simple replacement of `post` by `POST` in the expectation. Mind fixing that here too?

dblock:
Thanks

dblock:
Btw, I am pretty sure I had `Thread.pass` on purpose here and not a `join`, at least in EventMachine times. If you have a moment, do actually please test the EM code?

PR #209, opened 2018-06-15
Allow either attachments or text

peterzhu2118:
For `chat_postEphemeral`, either `text:` or `attachments:` is required, but the current code doesn't allow that (it checks that `text:` always exist).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,12 @@
 ### 0.11.2 (Next)
 
 * Your contribution here.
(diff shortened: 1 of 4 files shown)


dblock:
This file is automatically generated by lib/tasks/web.rake from https://github.com/slack-ruby/slack-api-ref, take a look at how that's done. You need to patch the API description similarly to what's in https://github.com/slack-ruby/slack-ruby-client/tree/master/lib/slack/web/api/patches.

peterzhu2118:
@dblock How does it look now? Should I update the tests as well for CI to pass?

dblock:
Yes please, changelog & whatever the bot complains about until it's green. Thanks.

peterzhu2118:
@dblock Done 😄! I've also updated the date in the changelog for version `0.11.1` because I think you meant 2018 rather than 2017.

dblock:
Merged, thank you.

Issue #223, opened 2018-09-07
Add websocket-level ping monitoring 


dblock:
Ping monitoring in slack-ruby-bot-server checks status of the bot, but https://github.com/faye/websocket-driver-ruby/issues/61#issuecomment-419256695 is saying that we could be sending a ping at websocket level. Which should cause the next read to return a `nil` and abort an offline situation.

ioquatix:
@dblock How are you going with this? I am back in NZ and can take a look at it if you still require assistance.

dblock:
I am not doing any work rn with this. The replacement of celluloid with async has shipped, but did nothing to improve the underlying problem.

RodneyU215:
I started to take a look at this last Friday and I believe I'll have a PR ready for review in a few hours.

RodneyU215:
@ioquatix think you'd be interested in reviewing and weighing in on my approach here #226?

dblock:
Implemented for async-websocket in https://github.com/slack-ruby/slack-ruby-client/pull/226.

dblock:
Implemented for Celluloid and EM in https://github.com/slack-ruby/slack-ruby-client/pull/236.

PR #226, opened 2018-09-25
Adding Slack Socket level ping/pong

RodneyU215:
###  Summary
This PR attempts to add websocket-level ping monitoring (#223):
- I've added a check to validate the socket is connected before sending any outgoing messages.
- I've implemented `heartbeat` following [Slack's RTM guidelines](https://api.slack.com/rtm). See "Ping and Pong".
- The argument in this statement `heartbeat 30` should likely be configurable.

It's not quite "shippable" yet, but I'd love some feedback on this.

### Requirements
* [x] I've read and understood the [Contributing Guidelines](https://github.com/slack-ruby/slack-ruby-client/blob/master/CONTRIBUTING.md) and have done my best effort to follow them.


dblock:
Looks like a solid start. Thanks!

- We need this for all RTM support (eventmachine, celluloid and async); we actually mostly care about async since it's supported and recommended today as default
- Refactor parts that can be shared such as storing ping/pong for the above, there's a lot of looping and some OOO could help :)
- There's a thread and updating a hash from multiple threads if I am not mistaken, is that thread safe?
- Is the ping thread dying on any kind of send error?

RodneyU215:
Thanks for the feedback @dblock!

I've been thinking of ways to refactor this as well as chatting with a few other engineers at Slack and I wanted to get your perspective on a new approach. 

Taking your feedback into consideration what do you think about implementing an `async` specific ping method in `Slack::RealTime::Concurrency::Async::Socket` that recursively sent out messages and checked to to ensure the connection was still “alive”.
 
```Ruby
          def ping(delay, task, ping_id=0)
            unless @alive
              disconnect! if connected?
              return close
            end

            ping_data = { type: 'ping', id: ping_id }
            send_data(ping_data.to_json)
            @alive = false

            task.sleep delay

            ping(delay, task, ping_id + 1)
          end
```
This would require us to update the `connect!` method so that by default each `pong` event would update the socket.
```
          def connect!
            super do |driver|
              driver.on :message do |event|
                event_data = JSON.parse(event.data)
                @alive = true if event_data['type'] == 'pong'
              end
              yield driver
            end
            run_loop
          end
```
Lastly we’d ensure that this was run on a different block from the client’s run_loop following the example [here](https://github.com/socketry/async/issues/27#issuecomment-421660836).
```
          def start_async(client)
            Thread.new do
              ::Async::Reactor.run do
                ::Async::Reactor.run do
                  client.run_loop
                end
                ::Async::Reactor.run do |task|
                  @alive = true
                  ping(task)
                end
              end
            end
          end
```

What do you think?

dblock:
Loos reasonable and I would merge support for just one of the async implementations for this.

I don't see it being difficult to generalize this to all 3 EM, celluloid, async, so I encourage you to implement this all the way for one, then expand from there.

lib/slack/real_time/socket.rb
```diff
@@ -81,6 +89,27 @@ def port
...
+        send_data(ping_data.to_json)
+        @alive = false
+
+        if block_given?
+          yield delay
+        else
+          sleep delay
+        end
+
+        ping(delay, ping_id + 1, &block)
```
dblock: This is going to blow up the stack eventually, won't it?

dblock:
@RodneyU215 This is a lot better.

- Make the ping loop iterative, not recursive.
- Ping loop isn't resilient, I am pretty sure that sending data over a dead socket will abort the ping loop while the connection may still be very much alive and just keep going.

lib/slack/real_time/concurrency/async.rb
```diff
@@ -15,7 +15,14 @@ class Socket < Slack::RealTime::Socket
           def start_async(client)
             Thread.new do
               ::Async::Reactor.run do
-                client.run_loop
+                ::Async::Reactor.run do
+                  client.run_loop
+                end
+                ::Async::Reactor.run do |task|
+                  ping do |delay|
```
dblock: Maybe model this as `run_loop` into a `run_ping` that's something that does ping every once in a while and recovers from errors.

RodneyU215: Can you expand on the types of errors you'd like to recover from as well as what the expected recover steps would be?

dblock: It can really be anything, but from looking at the code if `send_data` fails in any way the ping worker dies.

dblock:
I think you're getting close @RodneyU215, nice work. 

* Error handling is an important one. It also looks pretty trivial to make it work for all async drivers from here.
* Do we need a way to disable this in configuration? Configure frequency of ping?
* README should be talking about pings

RodneyU215:
Thanks for all of the feedback you've given! I'll spend a little more time on error handling Monday.

> Configure frequency of ping?

Upon closer inspection around the configuration it appears that there's already a option for `websocket_ping` which by default is set to `30`. I've switched the implementation to use this.

> Do we need a way to disable this in configuration?

I've added a check to see if `websocket_ping` was set. If users set this to `0` or `nil` then the code will not run.

> README should be talking about pings

The `websocket_ping` lists the default behavior in [#configuring-slackrealtimeclient](https://github.com/slack-ruby/slack-ruby-client/blob/master/README.md#configuring-slackrealtimeclient) as "The number of seconds that indicates how often the WebSocket should send ping frames, default is 30.". This to me implies that it was supposed to be behaving this way the whole time. Is there any other detail you wanted added? Seems like a solid description.

RodneyU215:
I'd like to talk a bit more around error handling. If the driver is unable to communicate with the server I can only really see reestablishing the connection or exiting the application as the best next steps.

In the current implementation on subsequent unanswered pings we disconnect and close the Socket connection. So that when `send_data` is called immediately after it'll throw a `Slack::RealTime::Socket::SocketNotConnectedError`.

To keep the application running and to reestablish the connection what do you think about catching that error, running `connect` again and retrying the ping task?
```Ruby
::Async::Reactor.run do |task|
  begin
    client.run_ping do |delay|
      task.sleep delay
    end
  rescue Slack::RealTime::Socket::SocketNotConnectedError
    connect
    retry
  end
end
```

dblock:
@RodneyU215 Sounds like a good approach. Since this will be the same for every driver, see how you can sink this implementation into a shared class.

dblock:
The websocket ping description is good enough. I guess before we were documenting it but it was actually never run :) I would change "should send frames" to describing what actually happens now with this code somewhere.

lib/slack/real_time/socket.rb
```diff
@@ -30,6 +34,13 @@ def connect!
         connect
         logger.debug("#{self.class}##{__method__}") { driver.class }
 
+        @alive = true
+
+        driver.on :message do |event|
+          event_data = JSON.parse(event.data)
```
dblock: I am not sure we do, but do we need to handle garbage from the server here?

RodneyU215: It's a good idea, though I'd recommend handling that in a separate PR. One where `Slack::Messages::Message` is used and updating it so that it would take care of any parsing errors that may occur.

RodneyU215:
Okay, so I've rebuilt things a bit to implement the ability for the client to automatically reconnect on ping failures. It felt a bit like scope creep but overall I think it's a better implementation. I've tested things manually and it worked, but I'll add unit tests tomorrow. Let me know what you think.

lib/slack/real_time/socket.rb
```diff
@@ -30,9 +35,46 @@ def connect!
...
+        end
+
+        Thread.new do
+          ping_int = 0
+          loop do
+            sleep delay
+            ping_data = { type: 'ping', time: Time.now, id: "p#{ping_int += 1}" }
+            ping(ping_data)
+          end
+        end
```
ioquatix: Where do you call `Thread#join`? If two threads try to write to the socket at the same time, what happens when the data becomes interleaved?

lib/slack/real_time/concurrency/async.rb
```diff
@@ -15,7 +15,22 @@ class Socket < Slack::RealTime::Socket
           def start_async(client)
             Thread.new do
               ::Async::Reactor.run do
```
ioquatix: You can simplify this code by something like 

```ruby
::Async::Reactor.run do |task|
  task.async do
    # nested task
  end
end
```

lib/slack/real_time/socket.rb
dblock: No worries. Please do make suggestions about this implementation! There's a lot of looping around in that calling code :)

ioquatix:
Okay, I looked at what you are doing. I think there is a fundamental issue with how this code is structured, but that being said, here is a rough outline of what I'd do:


```ruby
class Client
	attr :last_message_at
	
	def time_since_last_message
		Async::Clock.now - @last_message_at
	end
	
	def run_loop
		on(:message) do
			@last_message_at = Async::Clock.now
		end
	end
end

class Socket
	def start_async
		::Async::Reactor.run do |task|
			task.async do
				while @running # or true, some way to control this loop for graceful shutdown
					# Probably make this one method?
					client.build_socket
					begin
						client.run_loop
					rescue EOFError, EPIPE #, etc
						# Log and loop around again
					end
				end
			end

			task.async do |subtask|
				while true
					subtask.sleep delay
					
					client.ping # Will eventually trigger some kind of "pong".
					
					if client.time_since_last_message > (delay * 2)
						# Connection failed
						client.close
					end
				end
			end
		end
	end
end
```

ioquatix:
As a side note, you don't need to send the ping if some other message was received recently... you'd just dynamically adjust the delay, and then when you wake up, don't send ping if something else was sent in the mean time. That avoids sending unnecessary junk.

ioquatix:
Okay, so the fundamental issue is the callback structure of the code and basically it makes it _very_ hard to reason about. The main issue I have when reading the code is understanding the call flow and sequence of events that lead up to a particular line being executed. The solution that `async` brings to the table is ensuring your call graph is still essentially linear even with blocking operations. You still need to write sane, linear, readable code, but it's much harder to make mistakes and those mistakes are generally more obvious.

The next step is to consider invariants and how they are maintained. For example, are your connection stateful or stateless and where do you maintain that state. How is it affected by reconnections - is it modelled explicitly or implicitly (e.g. some kind of `on(:connect)`).

You propose reconnecting to the same URL, but is this valid? Is it possible after 24 hours, that URL changes? Where in your API do you need to implement the "timeout/reconnect" loop? That loop probably needs to be far up the call tree.

I see you try to restart the run loop in the exception handler, but then after that call retry. That code already seems wrong, do you want to `retry` the event loop after it's been gracefully shut down?

Ideally, you have some way to signal a disconnect from any part of the code and take appropriate action. If you are not careful, tracking state (which naturally includes which event handlers are registered) becomes difficult and buggy. Think about the following flow:

- Create client
- Register some events
- Loop connect/reconnect

After the first reconnect, the initial events would be lost because the entire web socket is thrown away, no? If things are stateful, it's even more tricky to handle correctly.

lib/slack/real_time/concurrency/async.rb
RodneyU215: Great point! I'll fix this.

RodneyU215:
I push that last commit before I had to chance to see all of this new feedback.

@ioquatix Thanks for the added insight! Dynamically adjusting the delay before sending out a ping based on the last message is a great suggestion. I also agree that using `task.async` is a lot cleaner. I'll definitely go back and update these.

In the code you suggested I noticed that you've chosen not to check for a `pong` event. Was this to minimize the amount of state needed to update `time_since_last_message` in your example? It feels a little odd to skip over any sort of validation that a `pong` message was sent back given that it validates the `ping`.

Other than that It sounds like the only thing left is the logic to restart the web socket connection via the new method `restart_async`. 
>I see you try to restart the run loop in the exception handler, but then after that call retry. That code already seems wrong, do you want to retry the event loop after it's been gracefully shut down?

In the latest version the code attempts to rebuild the connection via `build_socket` before running the event loop again in a new async block. Did you still feel as though this is wrong?

>You propose reconnecting to the same URL, but is this valid? Is it possible after 24 hours, that URL changes? Where in your API do you need to implement the "timeout/reconnect" loop? That loop probably needs to be far up the call tree.

WebSocket Message Server URL's are only valid for 30 seconds. But because we call `build_socket` when attempting to reconnect we retrieve and set a new URL.

lib/slack/real_time/socket.rb
```diff
@@ -59,6 +64,14 @@ def restart_async(_client)
         raise NotImplementedError, "Expected #{self.class} to implement #{__method__}."
       end
 
+      def time_since_last_message
+        current_time - @last_message_at
+      end
+
+      def current_time
+        Time.now
```
ioquatix: `Time.now` is not monotonic.

dblock: https://blog.dnsimple.com/2018/03/elapsed-time-with-ruby-the-right-way/

Running to work to fix all the instances of `Time.now` that I have all over the place :)

ioquatix:
> In the code you suggested I noticed that you've chosen not to check for a pong event. Was this to minimize the amount of state needed to update time_since_last_message in your example? It feels a little odd to skip over any sort of validation that a pong message was sent back given that it validates the ping was actually received.

It's really not necessary to confirm it's received. If no message is received within a certain timeframe, then the connection is considered dead. You don't care what kind of message it is, it doesn't change the outcome or behaviour.

> In the latest version I've push this logic into a method called restart_async. The code however always rebuilt the socket connection before running the event loop again in a new async task. Can you elaborate on why this seems wrong? I don't mean to be dense; I'm having a hard time seeing the problems this would cause.

It looks like that code path has been removed. I was referring to https://github.com/slack-ruby/slack-ruby-client/pull/226/commits/88ea7a5cb5b5899cf649244cf89f33f4bb4e94f3#diff-7de932a86b84fcf2a3e218dd5515f726L31

lib/slack/real_time/socket.rb
ioquatix: Haha, yeah, it's a common problem :p

ioquatix:
Ultimately, I think you just want to ensure all state is cleaned up correctly and that an entirely new instance is re-created. Because you are supporting different concurrency models, this is tricky. You are using a `Thread` as your single high level abstraction, but keep in mind you still want to ensure that any run-loop specific concerns are taken care of correctly and that you aren't leaking threads, file descriptors, etc.

lib/slack/real_time/socket.rb
RodneyU215: Solid feedback! Thanks for the link dblock! I'll change that.

RodneyU215:
@dblock I've addressed the outstanding feedback and added a few unit tests. The only thing I see to expand on is updating the `Celluloid` and `EventMachine` sockets so that their implementations of `start_async` kicks off the `run_ping` loop as well as implements the `restart_async` method. 

While I'd be happy to help with that; I believe it may be best for a future PR. This one has been open for a bit and I believe should help out a lot of people who may be currently experiencing disconnects with the clients default `Async` socket.

Would this work for you?

ioquatix:
You might like to add this to your specs: https://github.com/socketry/async-rspec/blob/master/lib/async/rspec/leaks.rb

It detects leaks. It's not async specific, it works with any spec that opens/closes IOs.

CHANGELOG.md
```diff
@@ -1,7 +1,11 @@
-### 0.13.2 (Next)
+### 0.13.3 (Next)
 
 * Your contribution here.
 
+### 0.13.2 (2018/10/16)
```
dblock: This hasn't shipped, so please put this back, we do this during release.

dblock:
> You might like to add this to your specs: https://github.com/socketry/async-rspec/blob/master/lib/async/rspec/leaks.rb
> 
> It detects leaks. It's not async specific, it works with any spec that opens/closes IOs.

Lets def add this in another PR!

dblock:
@ioquatix Appreciate another code review of the state of things here when you get a chance. Thank you!

CHANGELOG.md
RodneyU215: Whoops! I'll fix it now.

RodneyU215:
@dblock I'm not going anywhere! 😊I'd like to help get EM/Celluloid implemented too. We'll have more time to jump on it after this [tour around Europe](https://medium.com/slack-developer-blog/announcing-the-slack-platform-euro-tour-2018-29cd4abe7ef4).

dblock:
I've merged this, thanks for your time @RodneyU215. Lets try to get EM/Celluloid in soon, I'll spend some time upgrading one of my bots to this and see if this works in production, too.

dblock:
@RodneyU215 Take a look at the broken build in https://travis-ci.org/slack-ruby/slack-ruby-client/jobs/442629093? 

We run integration tests with a SLACK_API_TOKEN on master only, so this wasn't surfaced in this PR.

RodneyU215:
I'm looking into this now!

ioquatix:
It looks like at least one aspect is that you are missing `require 'async/clock'`.

dblock:
I've released:

- slack-ruby-client 0.14.0
- slack-ruby-bot 0.12.0
- slack-ruby-bot-server 0.9.0

Let's see how this puppy works and be attentive to fixing new bugs. Appreciate everyone's hard work!

dblock:
RT https://twitter.com/dblockdotorg/status/1100120314355302402

PR #232, opened 2018-10-18
Addressing a few issues with #run_ping! (Fixes #231)

### LINKED ISSUE #231: PR #226 broke integration tests
There appears to be a missing `async/clock` require statement as well as an issue where the client is not shutting down the reactor when stop is called.

I'm working to fix this now. WIP Branch: [RU_fixing_issue_231](https://github.com/slack-ruby/slack-ruby-client/compare/master...RodneyU215:RU_fixing_issue_231)

RodneyU215:
###  Summary
This PR addresses several issues with the initial implementation (#226) of `Slack::RealTime::Client#run_ping!`; a method which is responsible for ensuring the `Slack::RealTime::Client` stays connected until `Slack::RealTime::Client#stop!` is called.

- `Slack::RealTime::Client#close` will no longer assign nil to `@socket` because the reactor would still be running in the background if `websocket_ping` is set.
- `Slack::RealTime::Concurrency::Async::Socket#disconnect!` now cancels all timers (via `@reactor.cancel`). This shuts down the `websocket_ping` timer.
- I've removed the loop in `Slack::RealTime::Client#run_ping!` and now use on the builtin function `Async::Reactor#every` to run the ping on an interval.
- I've added `Slack::RealTime::Client#restart_sync` which is responsible for obtaining a new websocket url and attempting to connect to it.
- If the connection is closed and we've not called `Slack::RealTime::Client#stop!`; on the next `Slack::RealTime::Client#run_ping!` we'll attempt to restart the connection (via `Slack::RealTime::Client#restart_sync`)
- If we're still not connected, after attempting to restart the connection (via `Slack::RealTime::Client#restart_async`), we reraise `Slack::RealTime::Client::ClientNotStartedError`
- I've also added a new integration test to ensure it works as expected.

### Requirements
* [x] I've read and understood the [Contributing Guidelines](https://github.com/slack-ruby/slack-ruby-client/blob/master/CONTRIBUTING.md) and have done my best effort to follow them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### 0.13.2 (Next)
 
 * Your contribution here.
+* [#232](https://github.com/slack-ruby/slack-ruby-client/pull/232): Addressing a few issues with #run_ping! (fixes #231) - [@RodneyU215](https://github.com/RodneyU215).
 * [#226](https://github.com/slack-ruby/slack-ruby-client/pull/226): Added periodic ping that reconnects on failure for `async-websocket` - [@RodneyU215](https://github.com/RodneyU215), [@dblock](https://github.com/dblock), [@ioquatix](https://github.com/ioquatix).
 
 ### 0.13.1 (2018/9/30)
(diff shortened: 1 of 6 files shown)


RodneyU215:
#231

dblock:
Looks good.

PR #236, opened 2018-10-31
Implemented ping support for Celluloid and EventMachine.

dblock:
Closes https://github.com/slack-ruby/slack-ruby-client/issues/223

@RodneyU215 Want to code review this?


RodneyU215:
Celluloid and EventMachine need to ensure that the timers are canceled on disconnect. This prevents the connection from being reestablished when the user intentionally closes it. This was achieved for `async-websocket` by surfacing the `#cancel` method on `@timers`:
```Ruby
module Slack
  module RealTime
    module Concurrency
      module Async
        class Reactor < ::Async::Reactor
          def_delegators :@timers, :cancel
        end

        class Socket < Slack::RealTime::Socket
          def disconnect!
            super
            @reactor.cancel
          end
        end
      end
    end
  end
end
```

lib/slack/real_time/concurrency/celluloid.rb
```diff
@@ -73,7 +73,14 @@ def start_async(client)
           end
 
           def run_client_loop
+            current_actor.every @client.websocket_ping do
```
RodneyU215: I believe we want this wrapped in a conditional `if @client.run_ping?`.

RodneyU215:
We also have to implement a `#restart!` method for both Celluloid and EventMachine. This method is responsible for updating web socket url to the new one and kicking off the run_loop again. Here's initial approach we took in `async-websocket`. Note: there's going to be a snag around how `client` is passed.

```
module Slack
  module RealTime
    module Concurrency
      module Async
        class Socket < Slack::RealTime::Socket
          def restart_async(client, new_url)
            @url = new_url
            @last_message_at = current_time
            return unless @reactor
            @reactor.async do
              client.run_loop
            end
          end
        end
      end
    end
  end
end
```

lib/slack/real_time/concurrency/celluloid.rb
dblock: I implemented this with tests in https://github.com/slack-ruby/slack-ruby-client/pull/236/commits/019594953c372736172ce402d83b482888fa8921

dblock:
Made some progress, but definitely could use help with the other TODOs, especially with proper tests that fail in Celluloid/EM and pass in async rn.

dblock:
@RodneyU215 if you want to help with this one, that'd be greatly appreciated, I won't have time for it for a while

RodneyU215:
Thanks for the nudge! I'll work on it this week.

RodneyU215:
> Celluloid and EventMachine need to ensure that the timers are canceled on disconnect.

I tested this for EventMachine (via `faye-websocket`) and all timers are canceled when `EventMachine.stop` is called. I've not been able to find a comparable method in Celluloid however. Instead Celluloid continues to have an issue where the Actor crashes even after the main thread has been closed.

> We also have to implement a #restart! method for both Celluloid and EventMachine.

I've implemented the `#restart!` method for EventMachine. You can find that PR [here.](https://github.com/dblock/slack-ruby-client/pull/1) I've not had much success with Celluloid, but can make another attempt in a few days.

Side note: `celluloid-io` appears to be unmaintained do we still want to continue to support it?

dblock:
@RodneyU215 I think we should at least _try_ supporting celluloid. While unmaintained there's a slew of bots out there using it, including my own :)

I am going to merge this on green. How do I reproduce the problem with Celluloid?

dblock:
I merged this. Let's move from here.

RodneyU215:
@dblock Thanks for moving things forward. The issue with Celluloid surfaced by just running the integration tests. Starting tomorrow I'll have some free time over the next couple weeks to dig deeper into that issue.

dblock:
@RodneyU215 Could really use some help here so we can get the next release out, haven't had much time to deal with Celluloid and EM, but would be happy to test on some production bots. Integration tests seem to pass, so...

RodneyU215:
@dblock I believe EM is stable; I'm still working on verifying that Celluloid works as expected. I'm glad to hear that the integration tests are passing. This leads me to believe that it's likely an issue for me locally. 

I've got a few other things to get through at work before Friday but I'll continue working on it this weekend and follow up by Monday.

PR #239, opened 2018-12-04
handle nil attachments in a safe way

jmanian:
In our app we started receiving some `invalid_attachments_format` errors for certain workspaces, and I tracked it down to a few spots where we pass `attachments: nil` into `chat_postMessage`. I took a look in here to see what happens when you pass `attachments: nil` and the answer is that the JSON string for `attachments` is `null`. I was able to get around this error by simply omitting `attachments` instead.

We've actually have been doing this for a long time and it's never been a problem before, so I contacted Slack about it and they've identified it as a bug related to a beta feature for certain workspaces.

But there's really no good reason to ever send `null` for attachments, so I thought it might make sense to avoid that.

The change to `chat_update` is perhaps less self-explanatory than `chat_postMessage` and `chat_ephemeral`. Here's how `chat_update` behaves with `attachments`:

- omit `attachments`:
  - If there are existing attachments leave them alone
  - If there are no existing attachments don't add any
- pass `attachments` as `"[]"` or `null`:
  - Remove any existing attachments, if they exist
  - If there are no existing attachments don't add any

Anyway, I'm curious if you think this is an appropriate safeguard to add to this library, or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### 0.13.2 (Next)
 
 * Your contribution here.
+* [#239](https://github.com/slack-ruby/slack-ruby-client/pull/239): Automatically replace `nil` attachments with safer equivalents - [@jmanian](https://github.com/jmanian).
 * [#234](https://github.com/slack-ruby/slack-ruby-client/pull/234): Removing the immediate retry to give the client time to reconnect - [@RodneyU215](https://github.com/RodneyU215).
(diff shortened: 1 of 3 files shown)


jmanian:
I got word from Slack that the bug causing "null" attachments to yield `invalid_attachments_format` has been fixed. So now I'm not so sure if these changes are warranted. Maybe we should allow sending "null" via passing `nil`, in case some day that's meaningful.

dblock:
Want to just close this?

PR #244, opened 2019-01-14
Implementing #restart for the Celluloid Socket class.

RodneyU215:
I've also relocated the call to terminate the actor to the disconnect method. This finishes up the implementation so that Celluloid correctly handles restarts. See https://github.com/slack-ruby/slack-ruby-client/pull/236 for details.

Please hold off on merging though as I'm still working through testing these changes.

### DIFF
--- a/lib/slack/real_time/concurrency/celluloid.rb
+++ b/lib/slack/real_time/concurrency/celluloid.rb
@@ -59,7 +59,7 @@ module Slack
 
           def handle_read(buffer)
             logger.debug("#{self.class}##{__method__}") { buffer }
-            driver.parse buffer
+            driver.parse buffer if driver
           end
 
           def write(data)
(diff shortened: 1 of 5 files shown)


dblock:
Great, will wait for 🍏 too.

RodneyU215:
I've found a problem with my initial approach. I'm updating this PR today.

lib/slack/real_time/concurrency/celluloid.rb
```diff
@@ -37,12 +37,11 @@ def run_loop
...
-            begin
-              current_actor.terminate if current_actor.alive?
-            rescue StandardError
-              nil
-            end
+          end
+
+          def disconnect!
+            super
+            current_actor.terminate if current_actor.alive?
```
RodneyU215: This line of code would have never worked because you cannot terminate your current thread (i.e. `current_actor`). It consistently throws an error.
```Ruby
ThreadError:
       Target thread must not be current thread
```

spec/integration/integration_spec.rb
```diff
@@ -133,12 +133,29 @@ def stop_server
...
         @reply_to = data.reply_to
+        queue.push nil
         client.stop!
       end
       start_server
-      wait_for_server
+      queue.pop_with_timeout(5)
       expect(@reply_to).to be 1
     end
+    it 'rebuilds the websocket connection when dropped' do
```
RodneyU215: I'm not super happy with this test, but it's been the most consistent way I've been able to validate that the reconnect works as expected for each of the 3 supported concurrency libraries.

RodneyU215:
@dblock This PR is ready for a review.

dblock:
I merged this, thank you. Let's see how it behaves. I need to get a bot on this code, will try to do this next week.

PR #245, opened 2019-01-19
Added Slack::Events::Request#verify! to check whether an event request is coming from Slack.

### LINKED ISSUE #238: Does slack-ruby-client include a method I can use for verifying Slack event requests?
I've got the Events API working, and Slack is posting events to my server. I'm looking at the [Verifying Requests From Slack](https://api.slack.com/docs/verifying-requests-from-slack) documentation, and it doesn't mention Ruby under the "SDK support" section. Is that an oversight, or is there still no official way to verify Slack requests if you're using Ruby? I'm happy to follow the step-by-step instructions, but would prefer to use a library. Is there another Ruby gem that implements the same HMAC verification?

Thanks!

dblock:
Closes #238.

```ruby
slack_request = Slack::Events::Request.new(http_request)
slack_request.verify!
```


dblock:
@alexagranov @jmanian could I have your eyes on this? I'll keep adding to it for some basic events support. I think we can roll out the entire set of events as first class objects with time.

dblock:
cc: @ndbroadbent

dblock:
> Time.current

Since we `to_i` it, the time zone doesn't matter.

PR #261, opened 2019-04-05
Fix: handle driver#close E_PIPE.

### LINKED ISSUE #257: Occasional failures to reconnect
I have at least once instance where https://github.com/dblock/slack-sup disconnected a team and never reconnected it (nothing in logs). This is with all the fixes in #208 with 0.14.1. Restarting the bot helps obviously.

dblock:
Fixes #257 

I got lost in the async code to try to do this better. I don't think this is a fix, but maybe it's still something we should merge. Thoughts @ioquatix?


ioquatix:
This is not quite the right approach. I can make a PR.

dblock:
Closed in favor of https://github.com/slack-ruby/slack-ruby-client/pull/262

PR #268, opened 2019-05-09
Pin `async-websocket` to stable version.

ioquatix:
`v0.9.0` will contain breaking changes, but the updated interface should also be more stable.


dblock:
Thanks. Looking forward to upgrading whatever is needed here.

ioquatix:
I have completely rewritten the web socket implementation to be more scalable and I also believe it should be more reliable. Only time will tell I guess.

PR #273, opened 2019-06-03
Catch up slack API updates.

dblock:


dblock:
Replacing with https://github.com/slack-ruby/slack-ruby-client/pull/281

PR #291, opened 2019-07-28
Update README concerning OAuth and legacy tokens

### LINKED ISSUE #267: How to setup using OAuth rather than API key?
Hi, thanks for creating this gem!

Looking at the docs, I see instructions for getting started using an API Key. However, [reading the most recent docs on Slack, it seems this method has been deprecated](https://api.slack.com/custom-integrations/legacy-tokens). Instead, they now recommend creating "apps" which apparently by default use an OAuth flow.

Probably, you already know all this 🤷‍♂ 

**So my question:** How do I setup auth for this gem using the new, oauth workflow? For example, I would expect that would need to specify my `client_secret` and `client_id` in an initializer?

But I couldn't find any instructions in the readme. Any help would be much appreciated!

Thanks again!

jcraigk:
Fixes #267 

Related to https://github.com/slack-ruby/slack-ruby-bot-server/pull/105


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ### 0.14.4 (Next)
 
 * [#289](https://github.com/slack-ruby/slack-ruby-client/pull/289): Fix reconnects when ping timers under/overshoot - [@georgyangelov](https://github.com/georgyangelov).
+* [#291](https://github.com/slack-ruby/slack-ruby-client/pull/291): Update README concerning OAuth and legacy tokens - [@jcraigk](https://github.com/jcraigk).
```
dblock: No need for a CHANGELOG for docs, ignore danger complaining.

jcraigk: Just curious - why keep a manual CHANGELOG instead of auto-generating release notes from the commits?

PR #293, opened 2019-07-30
Rubocop auto-correct and comprehensive todo

jcraigk:
The goal here is to produce a more comprehensive Rubocop TODO that can lead to other PRs that address subsets of the list.  No functional changes are intended on the app or rspec logic.

 * Upgrade `rubocop` gem
 * Add `rubocop-rspec` and `rubocop-performance`
 * Start with a clean(er) `.ruboocop.yml`
 * Run `rubocop -a` to fix auto-correctable issues
 * Generate a fresh `.rubocop_todo.yml` using `rubocop --auto-gen-config`
 * Manually fix the `Metrics/LineLength` issues

Note the `to_s` on the `context`s in the specs is to allow using `described_class` which is a Rubocop auto-correctable offense that was breaking the specs in question.  That is potentially a Rubocop bug in that it shouldn't auto-correct issues that could potentially break logic.


dblock:
Thanks.

Can we please get rid of `# frozen_string_literal: true` changes? Exclude them in rubocop.yml. It's just noise.

lib/slack/web/api/endpoints/chat.rb
```diff
@@ -221,7 +223,7 @@ def chat_postMessage(options = {})
           #   Pass false to disable unfurling of media content.
           # @see https://api.slack.com/methods/chat.scheduleMessage
           # @see https://github.com/slack-ruby/slack-api-ref/blob/master/methods/chat/chat.scheduleMessage.json
-          def chat_scheduleMessage(options = {})
+          def chat_scheduleMessage(options = {}) # rubocop:disable Naming/MethodName
```
dblock: These are auto-generated files, so this gets wiped up next time they are generated. You should put back disabling this cop in rubocop.yml.

.rubocop.yml
```diff
@@ -1,21 +1,13 @@
...
+  - rubocop-performance
+  - rubocop-rspec
+
 AllCops:
+  DisplayCopNames: true
   Exclude:
     - vendor/**/*
     - lib/slack/web/api/slack-api-ref/**/*
-
-Metrics:
```
dblock: We basically ignore metrics everywhere, so maybe just leave this?

lib/slack/web/api/endpoints/chat.rb
jcraigk: Okay I didn't realize these were auto-generated.  If the rule exception were put back into`.rubocop.yml` then it would allow method naming violations in other areas of the code.  Do we want that?

Obviously as the maintainer of the project it's your call on what Rubocop rules are included. 
 My take is if we want the code to be more readable and performant, we should try to stick to a nearly empty `.rubocop.yml`.

.rubocop.yml
jcraigk: Why ignore Metrics?

jcraigk:
@dblock Concerning `frozen_string_literal: true`, when you say "just noise" - have you confirmed that it does not produce a performance boost in the app?  The basic assumption is that including the line in all files speeds up the execution of any Ruby program that uses strings (https://www.mikeperham.com/2018/02/28/ruby-optimization-with-one-magic-comment/).  I personally see `frozen_string_literal` used in all projects at my fulltime job as well as client work and I use it in my personal projects as well.  I know many senior Rubyists who do not consider it just noise (although I admit it is an annoying language feature to look at), so just curious why that sentiment is being expressed here.  That said I do disable `Layout/EmptyLineAfterMagicComment` so it just takes up one extra line at the top of the file.

lib/slack/web/api/endpoints/chat.rb
dblock: I agree in principle but also at some point there's signal vs. noise... anyway that's a philosophical debate.

You could modify the templates or exclude generated files as a solution if you want that. But either way modifying generated files isn't going to work.

.rubocop.yml
dblock: Personally I think they are noise. Line length is one of those things from the 80s :) But I am also happy if we don't ignore them as long as the violations are fixed.

dblock:
Regarding frozen string literals, it's free performance improvement indeed. I don't have any good arguments other than the ugly comment on top. I'm fine with it I guess, maybe it's time for me to just let that one go :)

From a more pragmatic POV for a library like this, I challenge anyone to show any performance gain at all in any real scenario. This is a direct proxy to Slack making network calls which take many magnitudes more time than anything you will ever gain from frozen string literals. But hey, free improvement "somewhere", right?

.rubocop.yml
jcraigk: As someone with a disability who is often forced to use limited screen real estate (an iPad hovering over my head as I lay on my back), I especially appreciate adherence to lines that do not exceed ~100 characters.  Also considering the ubiquity of web-based code reviews through github/gitlab where those systems impose their own visual limitations on the width available for reading the lines in a pull request, I think shorter lines are generally better (scrolling horizontally is generally a poor UX in this context).  Such limitations can be overridden with browser extensions, but I would argue that it's best to try to adhere to code that is readable in these contexts to make it inclusive and easy for everyone.  Additionally, I would argue that at least most English readers appreciate shorter lines when trying to comprehend code they have not encountered before (perhaps this is only my personal perception, however).

There are several other important Cops in the Metrics group including ClassLength and BlockLength.  I admit I usually override these, often doubling them to 200-250 depending on project.  BlockLength in rspec test is especially troublesome.  However, I would argue that any file/class/block exceeding about 350 lines becomes troublesome to navigate in modern development flows.

I wanted to raise these concerns before proceeding with ignoring Metrics altogether, but am happy to do so if that is what's preferred here.  Please advise, thanks!

jcraigk:
@dblock Yeah I hear you on the assertion that the performance boost of `frozen_string_literal: true` is probably negligible in most situations, but if we can get a performance boost in Ruby, I say go for it, especially when many Rubyists have just accepted the annoying line at the top at this point. I've been faced with having to give up my own preferences as I've adopted Rubocop more and more over the last few years, but ultimately it's really improved my ability to read others' code and write code that others can comprehend quickly.  I'll go ahead and apply `Layout/EmptyLineAfterMagicComment` so it just takes up a single line and see if that improves the perception of the noise a bit.

.rubocop.yml
dblock: Thanks @jcraigk for the long explanation.

First, I am sorry that I am, clearly, not sensitive enough to developers with disabilities such as yourself, that's my bad, on me.

I absolutely want your valuable contributions to this and other projects, so let's make it as easy as possible moving forward and not ignore metrics. Please adjust them as you see fit and we can discuss each individual one and I'll accept anything reasonable.

dblock:
I am OK merging this as is.

You should add a line to CHANGELOG, this is important work.

Shall we also change the templates for auto-generated code to include frozen string literals? Run `rake slack:api:update`, then `git add .` and run rubocop, it should not modify/flag any new files.

jcraigk:
@dblock Thanks for being so accommodating on this.  I'll do my best to followup with style fixes to reduce the `rubocop_todo.yml` list.

FYI I regenerated the TODO list with `rubocop --auto-gen-config --auto-gen-only-exclude --exclude-limit 100` to cover all the cops and ensure that no rule is simply disabled if more than 100 files violate it (not the case in this specific repo it seems).

I added the `fronzen_string_literal: true` to the templates.  Just to double check, is the `lib/slack/web/api/slack-api-ref` correct now?  I haven't used submodules in years so I'm not sure if I messed something up there.  Also I committed the new auto-generated files but let me know if those should just be omitted.

dblock:
I checked this out in a branch, rebased against master and did a `git submodule update`. 

Ran `rake slack:api:update` and that was clean. So we're good to go, merging.

PR #297, opened 2019-08-21
Cleanup various Rubocop offenses

jcraigk:
Fixes the following Rubocop violations:
 * Lint/AssignmentInCondition
 * Naming/FileName
 * Naming/MemoizedInstanceVariableName
 * RSpec/DescribeClass
 * RSpec/EmptyLineAfterExampleGroup
 * RSpec/EmptyLineAfterFinalLet
 * RSpec/ExpectInHook
 * RSpec/HookArgument
 * RSpec/NamedSubject
 * RSpec/NotToNot
 * RSpec/RepeatedDescription
 * RSpec/ScatteredLet
 * RSpec/ScatteredSetup
 * Security/Open
 * Style/AccessModifierDeclarations
 * Style/FrozenStringLiteralComment
 * Style/GuardClause
 * Style/ModuleFunction
 * Style/MultilineTernaryOperator
 * Style/NumericPredicate

Things to double check:
 * Removed what I think is redundant `slack-ruby-client.rb` file.  The `gemspec` defines the name of the gem (which includes the `-`s) and `snake_case` is preferred for Ruby filenames.
 * Definition of `open` method in `lib/slack/real_time/client.rb` was confusing Rubocop (thought it was `Kernel#open`) so I renamed the method.


lib/slack/config.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 module Slack
   module Config
-    extend self
+    extend self # rubocop:disable Style/ModuleFunction
```
dblock: Let's put this in the .yml as an explicit global exclusion?

lib/slack/real_time/client.rb
```diff
@@ -205,7 +205,7 @@ def send_json(data)
         @socket.send_data(data.to_json)
       end
 
-      def open(_event); end
+      def open_event(_event); end
```
dblock: Could somebody be relying on this behavior?

jcraigk: My naive take was that the author was not aware of `Kernel#open` and so didn't realize they were overriding a globally available method (I wasn't personally aware of the raw `open` either until this cop complained at me).  It looks like the method def was added here: https://github.com/slack-ruby/slack-ruby-client/commit/e7aaab2e3bdc6e15ae0f470b5f56d776f093d02d#diff-47933b59ae5516f473232936f333cbedR50 Do you have any insight on the intention there?

dblock: I doubt that I was intentional about it. It's fine to change it, but needs to be called out in UPGRADING and we should do a major version bump.

jcraigk: How does this warrant a version bump?  The functionality is exactly the same - we just rename the method so it doesn't override `Kernel#open`.  It's functionality equivalent, right?

jcraigk:
@dblock You're right, I guess we do need to keep the `ruby-slack-ruby.rb` file.  It's discouraged based on https://guides.rubygems.org/name-your-gem/ but I see a lot of gems violate this, including `rubocop-rspec`. 🤷‍♂

dblock:
I merged this.

lib/slack-ruby-client.rb
```diff
@@ -15,12 +15,12 @@
...
 begin
   require 'picky'
-rescue LoadError
+rescue LoadError # rubocop:disable Lint/HandleExceptions
   # ignore, only used in users_search
 end
 begin
   require 'openssl'
-rescue LoadError
+rescue LoadError # rubocop:disable Lint/HandleExceptions
```
booch: @jcraigk You could either enable the `AllowComments` option on the `Lint/HandleExceptions` cop, or just add a `nil` in front of the comments in the blank exception handler, instead of disabling the cop here.

lib/slack/real_time/client.rb
booch: It appears that the `open` method is here for subclasses to override (the template method pattern). I doubt that many user of the gem would subclass Client, and not many would have a reason to be notified of the open event. So it's probably OK not to bump the version.

lib/slack-ruby-client.rb
jcraigk: Ah good call.  It seems like the `nil` addition should produce the same results in Rubocop since it's functionally equivalent (the method returns `nil` implicitly on `rescue`)...I wonder if that should be changed on the Rubocop side... 🤔

PR #300, opened 2019-10-05
Update links in README to point to slack-ruby instead of dblock fork

kstole:
I just noticed it seemed a bit odd that some of the links go to the fork instead of the main repo


dblock:
Obviously this didn't break the build, but would be nice if someone looked into it.

PR #303, opened 2019-10-24
Adds views.publish method

jwright:
This PR adds the [views.publish](https://api.slack.com/methods/views.publish) method from the Slack API. This new method allows developers to post to the App Home using Block Kit.

By running `rake slack:api:update`, it also added the [oauth.v2.access](https://api.slack.com/methods/oauth.v2.access) method. This method does not have any complex options, so there was no custom changes needed here.

A general note, I would of liked to of refactored the `views.*` methods for serializing the view with a private method like the one below in order to remove the duplication of serializing the view option but running `rake slack:api:update` did not allow that. I am unsure if this possible but let me know if it is somehow and I will refactor it with the following:

```
# lib/slack/web/api/endpoints/views.rb

def views_publish(options = {})
  # ...
  options = serialize_view_option(options)
  post('views.publish', options)
end

# ...

private

def serialize_view_option(options)
  return options unless options.key?(:view)

  view = options[:view]
  view = JSON.dump(view) unless view.is_a?(String)
  options.merge(view: view)
end
```


dblock:
Regarding views, you can modify templates and do patches through this process.

PR #305, opened 2019-12-03
add admin web api methods

manuelmeurer:
I updated the API methods via `rake slack:api:update`, which added lots of new admin methods.
Once those are merged, we can remove the undocumented [`Slack::Web::Api::Endpoints::UsersAdmin.users_admin_invite`](https://github.com/slack-ruby/slack-ruby-client/blob/7c93cec20b87b38388b8afe092ca5d6adf71e51e/lib/slack/web/api/endpoints/users_admin.rb#L27)


manuelmeurer:
I changed a few small things in the changelog:
* the last entry referenced a wrong PR number (302 instead of 303) and misspelled the added endpoint (`views.published` instead of `views.publish`)
* most of the entries used past tense ("Added") and only a few present tense ("Add"). I changed the latter ones to past tense for consistency.

dblock:
This looks good, but there seem to be legit spec failures? I haven't looked any deeper than that.

manuelmeurer:
Ok, I'll investigate!

Why does Ruby 2.4.1 on Travis only run `bundle exec danger`, not the actual specs?
https://travis-ci.org/slack-ruby/slack-ruby-client/jobs/620311621

manuelmeurer:
I get the same failures locally when I run the specs on master, so I don't think they are related to the changes in this PR.

dblock:
> Ok, I'll investigate!
> 
> Why does Ruby 2.4.1 on Travis only run `bundle exec danger`, not the actual specs?
> https://travis-ci.org/slack-ruby/slack-ruby-client/jobs/620311621

Danger is what gives us changelog checking and stuff like that.

dblock:
> I get the same failures locally when I run the specs on master, so I don't think they are related to the changes in this PR.
> I'll try to find out what's going on anyways. :)

Yes thank you, we want that fixed first.

manuelmeurer:
@dblock The culprit is Faraday. They are preparing for a v1.0 release and (unintentionally?) introduced some breaking changes from 0.17.0 to 0.17.1:
https://github.com/lostisland/faraday/compare/v0.17.0..v0.17.1#diff-1b35e95f85a819bcd3c16d920ecdcffe

Basically, in 0.17.0, `Faraday::Error` inherits from `StandardError`, [so we have to add the possibility to initialize a `SlackError` with a response.](https://github.com/slack-ruby/slack-ruby-client/blob/v0.14.4/lib/slack/web/api/errors/slack_error.rb)

But in 0.17.1, [`Faraday::Error` already does this iself.](https://github.com/lostisland/faraday/blob/v0.17.1/lib/faraday/error.rb#L11-L24)

What to do? I added a check for the arity of `Faraday::Error#initialize` but I think it's ugly and brittle.
Alternatively, we could limit Faraday to either < 0.17.0 or 0.17.1.

dblock:
Are they releasing 0.17.2 fixing the regression? Otherwise make our code compatible with 0.17.1 and add a `>= 0.17.1` rather than limit below. We want to be compatible with the latest.

manuelmeurer:
I opened an [issue in the Faraday](https://github.com/lostisland/faraday/issues/1090) repo and will add a limit to `>= 0.17.1` for now.

manuelmeurer:
@dblock After the Faraday issue was fixed in https://github.com/slack-ruby/slack-ruby-client/pull/309, this should be good to go!

PR #307, opened 2019-12-09
Raise Slack-scoped exception if response not JSON

Deconstrained:
This will raise an appropriately scoped exception with a descriptive message if the Slack API responds with non-JSON, i.e. an nginx 502/500 HTML error page.

### DIFF
--- a/lib/slack/web/faraday/response/raise_error.rb
+++ b/lib/slack/web/faraday/response/raise_error.rb
@@ -6,6 +6,7 @@ module Slack
         class RaiseError < ::Faraday::Response::Middleware
           def on_complete(env)
             raise Slack::Web::Api::Errors::TooManyRequestsError, env.response if env.status == 429
+            raise Slack::Web::Api::Errors::SlackError.new("Invalid JSON received from Slack API.", env.response) if !body.is_a?(Hash)
             return unless (body = env.body) && !body['ok']
 
             error_message =
--- a/spec/slack/web/faraday/response/raise_error_spec.rb
+++ b/spec/slack/web/faraday/response/raise_error_spec.rb
@@ -57,5 +57,15 @@ RSpec.describe Slack::Web::Faraday::Response::RaiseError do
         )
       end
     end
+
+    context 'with non-JSON response body' do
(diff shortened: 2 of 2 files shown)


lib/slack/web/faraday/response/raise_error.rb
```diff
@@ -6,6 +6,7 @@ module Response
         class RaiseError < ::Faraday::Response::Middleware
           def on_complete(env)
             raise Slack::Web::Api::Errors::TooManyRequestsError, env.response if env.status == 429
+            raise Slack::Web::Api::Errors::SlackError.new("Invalid JSON received from Slack API.", env.response) if !body.is_a?(Hash)
```
dblock: The point here is not that we got something other than JSON, but that we got back a 500. Slack API is predictable as in with a 500 you don't get JSON. So you should specialize the error similarly to `TooManyRequestsError`, and check for the error code instead of body being a hash to avoid future surprises.

spec/slack/web/faraday/response/raise_error_spec.rb
```diff
@@ -57,5 +57,15 @@
         )
       end
     end
+
+    context 'with non-JSON response body' do
```
dblock: This should just be status 500 with a non-JSON body and we should get the expected error.

dblock:
I'm still interested in this, want to try and finish it @Deconstrained ?

dblock:
@agrobbin does https://github.com/slack-ruby/slack-ruby-client/pull/359 fully cover this one?

agrobbin:
@dblock I believe #359 combined w/ #350 handles this situation!

status | body | result
------ | ----- | ------
non-5xx | JSON | success
non-5xx | non-JSON | `Slack::Web::Api::Errors::ParsingError` from #350 
5xx | JSON | `Slack::Web::Api::Errors::UnavailableError` from #350 
5xx | non-JSON | `Slack::Web::Api::Errors::UnavailableError` from #359

dblock:
Thanks for the chart @agrobbin, super clear. Closing this.

PR #308, opened 2019-12-19
fix inheriting from Faraday::Error in Faraday v0.17.1

manuelmeurer:
Faraday changed the signature of an error class, as explained [here](https://github.com/slack-ruby/slack-ruby-client/pull/305#issuecomment-567227653).
This PR fixes inheriting from that error class and limits the Faraday version to `>= 0.17.1`.

I opened an [issue](https://github.com/lostisland/faraday/issues/1090) in the Faraday repo, let's wait for their response. Maybe they are open to reverting the change, then we could simple change the Faraday version requirement to `!= 0.17.1`.

### DIFF
--- a/lib/slack/web/api/errors/slack_error.rb
+++ b/lib/slack/web/api/errors/slack_error.rb
@@ -3,14 +3,7 @@ module Slack
   module Web
     module Api
       module Errors
-        class SlackError < ::Faraday::Error
-          attr_reader :response
-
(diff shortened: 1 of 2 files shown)


dblock:
Looks like for backwards compatibility we can assign `@response` after calling `super` on `Faraday::Error`.  I'm doing it in #309

manuelmeurer:
You're right, let's do that instead! 👍

PR #311, opened 2020-02-19
Add response_metadata to Web API errors

### LINKED ISSUE #310: RFC: Web API error response_metadata is buried
In some of the more recent Web API methods the most relevant error information is given in the `response_metadata`. For instance here's an example of Slack's response body for a bad call to `views.open`:
```json
{
  "ok": false,
  "error": "invalid_arguments",
  "response_metadata": {
    "messages": [
      "[ERROR] missing required field: title [json-pointer:/view]",
      "[ERROR] missing required field: blocks [json-pointer:/view]",
      "[ERROR] missing required field: type [json-pointer:/view]"
    ]
  }
}
```

The `error` value of `invalid_arguments` is basically useless on its own — it's necessary to look at the `response_metadata` to understand the problem with the request.

Currently this library pulls out the `error` value and uses this for the `message` on `SlackError` ([here](https://github.com/slack-ruby/slack-ruby-client/blob/40cf30ca99df782de66704616edfac1ef9683622/lib/slack/web/faraday/response/raise_error.rb#L11)). It's possible to access the `reseponse_metadata` with `slack_error.response.body.response_metadata`. However my problem is that when an error is raised  my logs only show the `error`, because that is the `message` (and as far as I can tell, `raise` prints `Class: message`).

My goal is to get the `response_metadata` into my logs. Here's one idea that I have for a feature in this library that would achieve that:
1. Add a config option for verbose Web API errors (`verbose_web_api_errors`?).
1. If this config option is set then the `message` on `SlackError` will be the entire body as JSON, rather than just the `error`.
1. The option would default to false so as to not break anyone's code.
1. `SlackError` would also gain a new attribute, perhaps `error`, which would be the value of `error` from Slack's response, so that this is easily accessed even when the verbose option is being used.
1. For completeness, `SlackError` would also gain a new attribute `response_metadata` containing just the `response_metadata` (as a hash-like object).

What do you think of this approach? I don't know much about logging, so maybe there's a better approach to this, either within this library or simply in my own app.

jmanian:
Relates to #310 

1. `SlackError#error` and `SlackError#response_metadata` methods give direct access to these fields in Slack's response body.
1. Improvements to README

-----

Original description below, from the first draft of this PR that did a lot more:

There are two main things happening here:
1. `SlackError#error` and `SlackError#response_metadata` methods give direct access to these fields in Slack's response body.
1. A new `verbose_errors` option changes the `message` on `SlackError` to include `response_metadata`.

I hit one snag, which is that I wasn't sure how to access the client instance within `Slack::Web::Faraday::Response::RaiseError`, which is where the error message is currently assembled. This means that I could not read the `verbose_errors` setting off of the client itself, but instead had to pull it from the config, which means that it is not configurable at the client instance level.

### DIFF
--- a/lib/slack/web/api/errors/slack_error.rb
+++ b/lib/slack/web/api/errors/slack_error.rb
@@ -10,6 +10,18 @@ module Slack
             super message
             @response = response
           end
+
+          def error
+            response.body.error
+          end
+
+          def errors
+            response.body.errors
+          end
+
+          def response_metadata
+            response.body.response_metadata
+          end
         end
       end
     end
--- a/lib/slack/web/client.rb
+++ b/lib/slack/web/client.rb
@@ -10,7 +10,7 @@ module Slack
 
       def initialize(options = {})
         Slack::Web::Config::ATTRIBUTES.each do |key|
-          send("#{key}=", options[key] || Slack::Web.config.send(key))
+          send("#{key}=", options.fetch(key, Slack::Web.config.send(key)))
         end
         @logger ||= Slack::Config.logger || Slack::Logger.default
         @token ||= Slack.config.token
--- a/lib/slack/web/faraday/response/raise_error.rb
+++ b/lib/slack/web/faraday/response/raise_error.rb
@@ -10,6 +10,11 @@ module Slack
 
             error_message =
               body['error'] || body['errors'].map { |message| message['error'] }.join(',')
+
+            if Slack::Web.config.verbose_errors && body['response_metadata']
+              error_message = "#{error_message}; #{body['response_metadata'].to_json}"
+            end
+
             raise Slack::Web::Api::Errors::SlackError.new(error_message, env.response)
           end
         end
--- a/lib/slack/web/faraday/response/raise_error.rb
+++ b/lib/slack/web/faraday/response/raise_error.rb
@@ -10,11 +10,6 @@ module Slack
 
             error_message =
               body['error'] || body['errors'].map { |message| message['error'] }.join(',')
-
-            if Slack::Web.config.verbose_errors && body['response_metadata']
-              error_message = "#{error_message}; #{body['response_metadata'].to_json}"
-            end
-
             raise Slack::Web::Api::Errors::SlackError.new(error_message, env.response)
           end
         end
--- a/spec/slack/web/api/errors/slack_error_spec.rb
+++ b/spec/slack/web/api/errors/slack_error_spec.rb
@@ -11,6 +11,7 @@ RSpec.describe Slack::Web::Api::Errors::SlackError do
     rescue described_class => e
       expect(e.response).not_to be_nil
       expect(e.response.status).to eq 200
+      expect(e.message).to eql 'invalid_arguments'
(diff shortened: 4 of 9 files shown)


lib/slack/web/api/errors/slack_error.rb
```diff
@@ -10,6 +10,18 @@ def initialize(message, response = nil)
             super message
             @response = response
           end
+
+          def error
```
jmanian: This is sort of a horrible name. I used it because it's the name of the field in the response body, but I'm open to changing it. One alternative is `error_code` (and `error_codes` below).

lib/slack/web/client.rb
```diff
@@ -10,7 +10,7 @@ class Client
 
       def initialize(options = {})
         Slack::Web::Config::ATTRIBUTES.each do |key|
-          send("#{key}=", options[key] || Slack::Web.config.send(key))
+          send("#{key}=", options.fetch(key, Slack::Web.config.send(key)))
```
jmanian: This turned out to not be directly relevant, but I noticed that if there was a truthy value in the config that it could not be overridden with a falsey value.

lib/slack/web/faraday/response/raise_error.rb
```diff
@@ -10,6 +10,11 @@ def on_complete(env)
 
             error_message =
               body['error'] || body['errors'].map { |message| message['error'] }.join(',')
+
+            if Slack::Web.config.verbose_errors && body['response_metadata']
```
jmanian: This is where I could not figure out how to access the client that was used to make the request. If there's a convenient way to do that then I can change this to pull the setting from the client.

dblock:
To be honest, I dislike this. We are introducing a global flag to control formatting of an error message. It just feels wrong to me.

Can we split this PR into two parts and get the non-controversial parts merged? I like `SlackError#error` and `SlackError#response_metadata` methods that give direct access to these fields in Slack's response body and much of the doc changes. We can think of how to solve your other problem then (I have some ideas)?

jmanian:
Sure, I see what you're saying. I'll pull out the global option and the corresponding change to the error message.

jmanian:
@dblock updated

spec/slack/web/api/errors/slack_error_spec.rb
```diff
@@ -4,13 +4,22 @@
 RSpec.describe Slack::Web::Api::Errors::SlackError do
   let(:client) { Slack::Web::Client.new }
 
-  it 'provides access to the response object', vcr: { cassette_name: 'web/auth_test_error' } do
```
dblock: I would also leave that old test in-place and add checks for its `.message`, `.error`, `.response_metadata`, it's a simpler case, but will prevent breaking it in the future.

dblock:
LGTM, feel free to (re)add that missing test and squash-merge on 💚

dblock:
Needs `rubocop -a ; rubocop --auto-gen-config`

jmanian:
Oh I just saw your comment. I resolved by following the rule.

Issue #204, opened 2018-04-10
Replace Slack API Ref with official OpenAPI Spec


Roach:
https://github.com/slackapi/slack-api-specs

jmanian:
Unfortunately the OpenAPI spec is not actively maintained at the moment and has gotten fairly out of date. I think we need to hold on this until the spec is being kept up to date more regularly, or else we won't have access to the newest methods.

I [chatted](https://dev4slack.slack.com/archives/C0K8617D5/p1584114323343900) with @episod in the Slack dev hangout and there's no concrete timeline of when it will be resurrected.

dblock:
Ugh. Sounds familiar. So we continue maintaining a community https://github.com/slack-ruby/slack-api-ref that can be kept up-to-date as needed by the consumers.

PR #312, opened 2020-03-11
add specific error classes for Web API

jmanian:
relates to #310

See comments below for some notes on the approach.


lib/tasks/web.rake
```diff
@@ -74,6 +74,33 @@ namespace :slack do
...
+          end
+        end.flatten.compact.uniq.sort
+
+        errors -= known_typos
+
+        error_template = Erubis::Eruby.new(File.read('lib/slack/web/api/templates/error.erb'))
+        errors.each do |error|
+          rendered_error = error_template.result(error: error)
+          File.write("lib/slack/web/api/errors/#{error}.rb", rendered_error)
+        end
```
jmanian: I considered putting all the classes into one file, instead of each class in its own file, since there's so many of them. I went this way because it follows the existing pattern more closely, but I can switch to one file if you think that's better.

lib/tasks/web.rake
```diff
@@ -74,6 +74,33 @@ namespace :slack do
...
+
+        # Use the official spec to pull a list of errors and create classes for each
+
+        spec = JSON.load(open('https://raw.githubusercontent.com/slackapi/slack-api-specs/master/web-api/slack_web_openapi_v2.json'))
+        known_typos = %w[account_inactiv token_revokedno_permission]
+
+        errors = spec['paths'].map do |_, path|
+          path.map do |_, v|
+            v['responses'].map do |_, response|
+              response.dig('schema', 'properties', 'error', 'enum')
```
jmanian: Is it OK to use `dig`? Can we assume anyone running this is on a new enough ruby? I'm not sure all the versions supported by the gem are new enough for `dig`.

lib/slack/web/faraday/response/raise_error.rb
```diff
@@ -10,7 +10,10 @@ def on_complete(env)
 
             error_message =
               body['error'] || body['errors'].map { |message| message['error'] }.join(',')
-            raise Slack::Web::Api::Errors::SlackError.new(error_message, env.response)
+
+            error_class = Slack::Web::Api::Errors::ERROR_CLASSES[error_message]
```
jmanian: My original plan was to use `constantize` to turn the error message into the corresponding error class, but then I came across [this (old) post](http://gavinmiller.io/2016/the-safesty-way-to-constantize/) which claims `constantize` can be a security risk, so I took its advice and built a hash to map from the message to the class.

lib/tasks/web.rake
jmanian: Actually it looks like we test back to 2.3, which is when `dig` was added.

lib/tasks/web.rake
```diff
@@ -74,6 +74,33 @@ namespace :slack do
           'bin/commands.rb',
           commands_template.result(files: data.keys.map { |key| key.tr('.', '_') })
         )
+
+        ### Errors
```
jmanian: I started out with all of this new stuff in its own rake task, `web_errors.rake`, but then after living with that for a bit I thought it was wrong to separate. Again I could go either way here.

lib/tasks/web.rake
dblock: I think 1 file is better for this since they are all 1-liners. Sort them alphabetically to have a stable list.

lib/tasks/web.rake
dblock: I suppose it's fine.

lib/slack/web/faraday/response/raise_error.rb
dblock: This is a response from the server, so you can theoretically `.constantize`, but this will not be future proof. You'd have to capture handle new codes with a `rescue`. It's much better to use a map and not to rely on the exception for flow.

lib/tasks/web.rake
dblock: I'd split it, but NBD.

lib/tasks/web.rake
```diff
@@ -74,6 +74,33 @@ namespace :slack do
...
+
+        # Use the official spec to pull a list of errors and create classes for each
+
+        spec = JSON.parse(open('https://raw.githubusercontent.com/slackapi/slack-api-specs/master/web-api/slack_web_openapi_v2.json').read)
```
dblock: Do you think we can do https://github.com/slack-ruby/slack-ruby-client/issues/204 as well soon? If so I wonder whether we should be downloading and committing this JSON to have stable history against which version of the API the generator ran?

jmanian: I definitely thought about #204 but decided it was too much for me to bite off right now. But I can definitely put something in to write and commit the JSON. Are you thinking just a `.json` file somewhere in the repo that has the exact contents of the spec?

dblock: We bring slack-api-ref into lib/slack/web/api/slack-api-ref, maybe just lib/slack-api-specs?

lib/slack/web/faraday/response/raise_error.rb
jmanian: Yes, that was another reason I didn't use `constantize`, because I would need to `rescue` for the fallback.

lib/tasks/web_errors.rake
```diff
@@ -0,0 +1,35 @@
...
+require 'json-schema'
+require 'erubis'
+
+namespace :slack do
+  namespace :web do
+    namespace :api do
+      desc 'Update Web API errors from official spec'
+      task :update_errors do
+        spec = JSON.parse(open('https://raw.githubusercontent.com/slackapi/slack-api-specs/master/web-api/slack_web_openapi_v2.json').read)
+        known_typos = %w[
```
dblock: Oh wow. Open a ticket with Slack for this or an issue in their repo?

dblock:
I'm good with merging this! Thanks, this is awesome.

dblock:
Actually hold on. 

Given https://github.com/slack-ruby/slack-ruby-client/issues/204#issuecomment-598805219, do we want to make the part that scrapes web API a feature of https://github.com/slack-ruby/slack-api-ref? That creates a staging area for all kinds of nonsense where we scrape the slack docs, read unmaintained API references, etc.

lib/tasks/web_errors.rake
jmanian: I opened an issue (https://github.com/slackapi/slack-api-specs/issues/34), but I also just asked @episod about this [on Slack](https://dev4slack.slack.com/archives/C0K8617D5/p1584118044350200?thread_ts=1584114323.343900&cid=C0K8617D5). He thinks it's a typo in the underlying metadata that feeds into the spec, but that the typo would not exist in the actual API responses. I wasn't sure before whether it was the right choice to omit these, but I think this new information validates the omission.

jmanian:
🤦‍♂ I just noticed that https://github.com/slack-ruby/slack-api-ref _is_ scraping the errors for every method. I was sure I had looked for this and didn't see it, which is why I turned to the OpenAPI spec.

Maybe for now it makes sense to shelve the OpenAPI spec entirely, since it's not actively maintained, and instead just pull the errors from the existing https://github.com/slack-ruby/slack-api-ref. That way everything continues to come from the one source.

It should be a simple change. Thoughts?

dblock:
> 🤦‍♂ I just noticed that https://github.com/slack-ruby/slack-api-ref _is_ scraping the errors for every method. I was sure I had looked for this and didn't see it, which is why I turned to the OpenAPI spec.
> 
> Maybe for now it makes sense to shelve the OpenAPI spec entirely, since it's not actively maintained, and instead just pull the errors from the existing https://github.com/slack-ruby/slack-api-ref. That way everything continues to come from the one source.
> 
> It should be a simple change. Thoughts?

Yes, I think that's better. I just fixed the build there.

jmanian:
Awesome, I'll make the switch here!

jmanian:
The errors in the method data didn't require any cleanup, so it was straightforward to map the existing data to a list of errors in the rake task. See what you think.

lib/tasks/web_errors.rake
dblock: Yes of course, I'm just saying that we need to let Slack know, which you did. Thanks.

dblock:
LGTM!

PR #325, opened 2020-05-28
Remove activesupport dependency

ianks:
For non-rails shops, depending on AS can be a non-starter. Luckily, this gem only relied on two small AS features from what I can tell: `class_attribute`, `String#blank?` and one occasion of `ActiveSupport#HashWithIndifferentAccess`. It was easy to adjust and as a result, it should be easier to adopt.

Thank you for this gem, our team is very excited to start using it!


lib/tasks/web.rake
```diff
@@ -2,14 +2,15 @@
...
 require 'erubis'
-require 'active_support'
-require 'active_support/core_ext'
 
 namespace :slack do
   namespace :web do
     namespace :api do
       desc 'Update Web API.'
       task update: [:git_update] do
+        require 'active_support'
```
dblock, line 11: Should be a developement dependency for this, otherwise the task will fail anyway without the gem installed separately. Also, maybe code can be fixed not to need it either?

lib/slack/messages/message.rb
```diff
@@ -2,10 +2,6 @@
 module Slack
   module Messages
     class Message < Hashie::Mash
-      def presence
```
dblock, line 5: This is an actual Slack field. Aren't we breaking an API here? Should have had tests for it though either way.

dblock:
@ianks Want to finish this?

lib/slack/messages/message.rb
ianks: I checked this a few times, and all the specs are indeed passing, and behavior seems correct.

lib/tasks/web.rake
ianks: Since I don't know what this task is supposed to do, I do not feel comfortable refactoring it. But I will make it a development dependency :smile:

lib/slack/web/pagination/cursor.rb
```diff
@@ -32,14 +32,14 @@ def each
 
                 client.logger.debug("#{self.class}##{__method__}") { e.to_s }
                 retry_count += 1
-                sleep(e.retry_after.seconds)
+                sleep(e.retry_after)
```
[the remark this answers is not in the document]
ianks: `retry_after` is an integer, so calling `.seconds` was effectively a no-op.

ianks:
> I think we're missing a spec. This overrides Object#presence, so when there's a presence field in the result you get its value and not the return of that object. Either way, why remove it? I think you should put this code back, and if you care write a spec or open an issue and I will get to it.

@dblock All specs pass without the `presence` method, but when I attempt to define it myself, I cannot get the specs to pass regardless of whether I use `:presence` or `'presence'`. Depending on the context, things break. Could you could clone down to take a look?

UPGRADING.md
```diff
@@ -1,6 +1,19 @@
 Upgrading Slack-Ruby-Client
 ===========================
 
+### Upgrading to >= 0.16.0
+
+As of 0.16.0, we no longer depend on `activesupport` as a dependency. This means that if you were relying on `activesupport` functionality unintentionally, certain things in you code may break. To fix this, you have two options:
```
dblock, line 6: Let's say this simpler

As of 0.16.0 `slack-ruby-client` no longer requires `activesupport`. If you are using activesupport, add `gem 'activesupport' to your Gemfile.`.

UPGRADING.md
```diff
@@ -1,6 +1,19 @@
...
+
+1. Update your code to use pure Ruby, without ActiveSupport methods (*recommended*).
```
dblock, line 8: This is really none of our business in this library :)

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ### 0.15.0 (Next)
 
+* [#325](https://github.com/slack-ruby/slack-ruby-client/pull/325): Remove activesupport dependency - [@ianks](https://github.com/ianks).
+* [#326](https://github.com/slack-ruby/slack-ruby-client/pull/326): Update new methods - [@wasabigeek](https://github.com/wasabigeek).
```
dblock, line 4: This shouldn't be in this PR. remove and rebase.

UPGRADING.md
```diff
@@ -1,6 +1,19 @@
 Upgrading Slack-Ruby-Client
 ===========================
 
+### Upgrading to >= 0.16.0
```
dblock, line 4: This should be 0.15.0 AFAIK, not 0.16.0.

dblock:
Merged via https://github.com/slack-ruby/slack-ruby-client/commit/7494411d04fe3127c2809eae81180886bd1f014b

Issue #320, opened 2020-04-26
Compatibility with conversations API and deprecations on November 25th, 2020


adrianomedeirossantos:
Hi everyone,

First of all, this is a great project and it helped me to create bots in Ruby pretty quick 👍 

As you guys can see in https://api.slack.com/changelog/2020-01-deprecating-antecedents-to-the-conversations-api, the Slack team has marked all Web API methods in the channels, im, groups, and mpim namespaces as deprecated. The reccomendation is to use `conversations` methods.

Is there any plan to make the ruby client compatible with that?

dblock:
Would appreciate help with these, first by updating https://github.com/slack-ruby/slack-api-ref that can properly reflect any deprecated APIs, then warnings for deprecated APIs in slack-ruby-client and finally removal of those in versions after November 25th, 2020.

adrianomedeirossantos:
How can I proper mark APIs as deprecated? For example, this one https://github.com/slack-ruby/slack-api-ref/blob/master/methods/groups/groups.archive.json?

dblock:
> How can I proper mark APIs as deprecated? For example, this one https://github.com/slack-ruby/slack-api-ref/blob/master/methods/groups/groups.archive.json?

Start by introducing a "deprecated" field into it.  This file is auto-generated by [a rake task](https://github.com/slack-ruby/slack-api-ref/blob/master/tasks/lib/slack_api/methods_spider.rb) by scraping Slack web pages, so assuming the method is marked as deprecated on the web page you should be able to lift it from there and add `"deprecated" : true` into the JSON.

wasabigeek:
@dblock, was trying to run `rake slack:web:api:update` and running into some issues with "patches":
```
...
- patching lib/slack/web/api/patches/chat.1.text-attachments-required.patch
- patching lib/slack/web/api/patches/chat.2.attachments-json.patch
- patching lib/slack/web/api/patches/chat.3.update-attachments-support.patch
error: patch failed: lib/slack/web/api/endpoints/chat.rb:154
error: lib/slack/web/api/endpoints/chat.rb: patch does not apply
rake aborted!
failed to apply patch
...
```
I'm not the most familiar with git patches, what should I do in this case?

dblock:
The path will need to be updated. See https://github.com/slack-ruby/slack-ruby-client/blob/master/CONTRIBUTING.md#patching-slack-web-api to get started.

wasabigeek:
[...]

PR #326, opened 2020-06-03
Update new methods

wasabigeek:
Prereq to https://github.com/slack-ruby/slack-ruby-client/issues/320.


wasabigeek:
@dblock, not entirely sure if I've approached the patching (dd6b484) the right way, appreciate feedback. It was a bit painful to patch `chat.rb`, wondering if there's a better way - this is what I did:
- it seems the underlying generated code drifted since the last patch so all the existing patches failed, I had to regenerate it without applying the patch (commented out the patching code in the rake task)
- committed the new generated code
- manually copied the code from old patches into the generated code
- ran the `git diff` command as mentioned in the README
Seems like there aren't that many method patches, so hopefully I won't need to repeat this many times when we add the deprecation message.

There's also a bunch of new methods / commands that got added in 80a3962, I wasn't sure if you wanted to add them, it's an easy revert anyway.

lib/slack/real_time/client.rb
```diff
@@ -160,6 +160,10 @@ def to_s
         end
       end
 
+      def reconnected?
+        @reconnected ||= false
```
dblock, line 164: This seems unrelated, but regardless you don't want to have a side-effect on calling a property method of assigning anything. So this could be `!! @reconnected` to coerce it into a boolean, for example.

lib/slack/real_time/client.rb
```diff
@@ -170,6 +174,7 @@ def restart_async
         @url = data.url
         @store = @store_class.new(data) if @store_class
         @socket.restart_async(self, @url)
+        @reconnected = true
```
dblock, line 177: Assuming you need this (sure why not), since this can happen multiple times I would make this a timestamp, so `@reconnected_at = Time.now.utc`, then implement both `reconnected_at` and `reconnected?`. Either way separate PR please.

lib/slack/web/api/patches/chat.1.patch
```diff
@@ -24,18 +27,18 @@ index 54a7db1..c535bb5 100644
...
--            throw ArgumentError.new('Required arguments :text or :attachments missing') if options[:text].nil? && options[:attachments].nil?
+-            throw ArgumentError.new('Required arguments :text missing') if options[:text].nil?
 +            throw ArgumentError.new('Required arguments :text, :attachments or :blocks missing') if options[:text].nil? && options[:attachments].nil? && options[:blocks].nil?
-             # attachments must be passed as an encoded JSON string
-             if options.key?(:attachments)
-               attachments = options[:attachments]
-               attachments = JSON.dump(attachments) unless attachments.is_a?(String)
-               options = options.merge(attachments: attachments)
-             end
++            # attachments must be passed as an encoded JSON string
```
dblock, line 36: Those + signs look like a patch bug, this code won't run :)

wasabigeek: Possibly missing something here, but isn't this the output of the diff? I was able to run the rake task fine, so I don't think it causes an issue with the patching. Example output from the [older files](https://github.com/slack-ruby/slack-ruby-client/pull/326/files#diff-75d9a357dc1666af97093cba14986108L10) also has the `+` signs.

lib/slack/real_time/client.rb
wasabigeek: I think I might have left this by accident while looking at https://github.com/slack-ruby/slack-ruby-bot/pull/244 >_<

wasabigeek:
> Let's split up this PR into manageable things to review, maybe start by taking out the reconnected business?

I've kept only the chat changes for now!

lib/slack/web/api/patches/chat.1.patch
dblock: Bah, I don't know what I was thinking, this is a patch file. You're right.

So what happened here? Before we were removing the code that did `JSON.dump` and now we put it back? Seems like the new(er) version is correct.

dblock: Oh looks like multiple patches got merged into one. I get it, ignore me.

CONTRIBUTING.md
```diff
@@ -75,13 +75,31 @@ rake slack:api:update
...
+3. Generate a patch:
+    ```
+    git diff --no-color HEAD lib/slack/web/api/endpoints/chat.rb > lib/slack/web/api/patches/chat.1.patch
+    ```
+4. Run `rake slack:api:update` to ensure that the patch is cleanly applied. Implement a test for the added or modified functionality and commit the patch file.
+
+##### Resolving Patch Errors
+
+The auto-generated method files may drift overtime e.g. new arguments may be added or descriptions changed. Since previous patches were based on the older auto-generated files, git may be unable to apply them to the new files. Resolving them requires some good ol' splicing:
+1. Comment out the patching code in `lib/tasks/web.rake`:
```
dblock, line 92: TODO for later, we should add a rake task option that skips patching, so `rake slack:api:update --no-patch` or something like that.

dblock:
Merged, 👍

PR #329, opened 2020-06-19
Log when methods are deprecated

wasabigeek:
Follow-up to https://github.com/slack-ruby/slack-ruby-client/issues/320 - log when a method is / will be deprecated.

There are some new methods that were added as well from the latest `slack-api-ref`, I'll open a new PR.


dblock:
Add a spec similar to [undocumented methods](https://github.com/slack-ruby/slack-ruby-client/blob/225c71ecbd57667133de993550e1ac59ec004739/spec/slack/web/client_spec.rb#L197).

I think we have to decide whether the method *is* or *will* be deprecated. I would say "is". If the method stops working on a certain date we should extract that into slack-api-ref and add the message in the warning along the lines of "it will stop working on ...".

wasabigeek:
What do you think about linking to the specific slack API instead, or copying the entire message? The deprecation warning actually has two dates e.g. https://api.slack.com/methods/channels.list
```
This method is deprecated. It will stop functioning in February 2021 and will not work with newly created apps after June 10th, 2020. Learn more.
Please use these methods instead:
conversations.list
users.conversations
```

dblock:
> What do you think about linking to the specific slack API instead, or copying the entire message? The deprecation warning actually has two dates e.g. https://api.slack.com/methods/channels.list
> 
> ```
> This method is deprecated. It will stop functioning in February 2021 and will not work with newly created apps after June 10th, 2020. Learn more.
> Please use these methods instead:
> conversations.list
> users.conversations
> ```

Even better.

wasabigeek:
There are some new methods picked up with the updated submodule, I haven't committed them - barring failed tests, will open a separate PR for them.

spec/slack/web/client_spec.rb
```diff
@@ -206,6 +206,17 @@
       end
     end
 
+    context 'calling deprecated methods' do
+      let(:client) { described_class.new }
+
+      it 'produces a warning' do
+        expect(client.logger).to receive(:warn).with(/^channels.archive: This method is deprecated./)
```
dblock, line 213: Check for a list of alternate methods here.

dblock:
Needs a `rubocop -a ; rubocop --auto-gen-config`. Looks good!

.rubocop_todo.yml
```diff
@@ -95,13 +95,13 @@ RSpec/InstanceVariable:
...
+# Offense count: 77
 # Configuration parameters: .
 # SupportedStyles: have_received, receive
 RSpec/MessageSpies:
   EnforcedStyle: receive
 
-# Offense count: 94
+# Offense count: 95
 RSpec/MultipleExpectations:
   Max: 5
```
wasabigeek, line 106: Do you think we could ignore these? Couldn't quite figure out how to get them working nicely together, this was the furthest I got:
```
it 'produces a warning' do
        logger = instance_double(Logger)
        allow(logger).to receive(:warn)
        allow(client).to receive(:logger).and_return(logger)
        allow(client).to receive(:post)

        client.channels_archive(channel: 'test')
        expect(logger).to have_received(:warn).with(/
          ^channels\.archive:\ This\ method\ is\ deprecated
          .+
          Alternative\ methods:\ conversations\.archive\.
        /x)
      end
```
And then it gives me `Prefer receive for setting message expectations.` 😞

dblock: I think ignoring expectations is fine. I usually just `-a` everything and selectively fix if I feel like it.

dblock:
Merged, thank you.

PR #331, opened 2020-06-29
Use conversations_list instead of deprecated channels_list

### LINKED ISSUE #328: Method Deprecated error when using conversations_history
Run this simple program:

```
# frozen_string_literal: true
require 'slack-ruby-client'

Slack.configure do |config|
  config.token = ENV['SLACK_API_TOKEN']
  raise 'Missing ENV[SLACK_API_TOKEN]!' unless config.token
end

client = Slack::Web::Client.new
client.auth_test

client.conversations_history(channel: '#somechannel')

```
And it will produce this error:

```
/usr/bin/bundle exec /usr/bin/ruby /Users/lorenanderson/ruby_slack/test_slack.rb
Traceback (most recent call last):
	17: from /Users/foobar/ruby_slack/test_slack.rb:12:in `<main>'
	16: from /Library/Ruby/Gems/2.6.0/gems/slack-ruby-client-0.14.6/lib/slack/web/api/endpoints/conversations.rb:70:in `conversations_history'
	15: from /Library/Ruby/Gems/2.6.0/gems/slack-ruby-client-0.14.6/lib/slack/web/api/mixins/channels.id.rb:19:in `channels_id'
	14: from /Library/Ruby/Gems/2.6.0/gems/slack-ruby-client-0.14.6/lib/slack/web/api/mixins/ids.id.rb:12:in `id_for'
	13: from /Library/Ruby/Gems/2.6.0/gems/slack-ruby-client-0.14.6/lib/slack/web/api/mixins/channels.id.rb:20:in `block in channels_id'
	12: from /Library/Ruby/Gems/2.6.0/gems/slack-ruby-client-0.14.6/lib/slack/web/api/endpoints/channels.rb:165:in `channels_list'
	11: from /Library/Ruby/Gems/2.6.0/gems/slack-ruby-client-0.14.6/lib/slack/web/faraday/request.rb:11:in `post'
	10: from /Library/Ruby/Gems/2.6.0/gems/slack-ruby-client-0.14.6/lib/slack/web/faraday/request.rb:26:in `request'
	 9: from /Library/Ruby/Gems/2.6.0/gems/faraday-1.0.1/lib/faraday/connection.rb:279:in `post'
	 8: from /Library/Ruby/Gems/2.6.0/gems/faraday-1.0.1/lib/faraday/connection.rb:492:in `run_request'
	 7: from /Library/Ruby/Gems/2.6.0/gems/faraday-1.0.1/lib/faraday/rack_builder.rb:153:in `build_response'
	 6: from /Library/Ruby/Gems/2.6.0/gems/faraday-1.0.1/lib/faraday/request/multipart.rb:25:in `call'
	 5: from /Library/Ruby/Gems/2.6.0/gems/faraday-1.0.1/lib/faraday/request/url_encoded.rb:25:in `call'
	 4: from /Library/Ruby/Gems/2.6.0/gems/faraday-1.0.1/lib/faraday/response.rb:11:in `call'
	 3: from /Library/Ruby/Gems/2.6.0/gems/faraday-1.0.1/lib/faraday/response.rb:11:in `call'
	 2: from /Library/Ruby/Gems/2.6.0/gems/faraday-1.0.1/lib/faraday/response.rb:65:in `on_complete'
	 1: from /Library/Ruby/Gems/2.6.0/gems/faraday-1.0.1/lib/faraday/response.rb:12:in `block in call'
/Library/Ruby/Gems/2.6.0/gems/slack-ruby-client-0.14.6/lib/slack/web/faraday/response/raise_error.rb:16:in `on_complete': method_deprecated (Slack::Web::Api::Errors::SlackError)

```

The error is being produced by this API helper class:

```
# frozen_string_literal: true
require_relative 'ids.id'

module Slack
  module Web
    module Api
      module Mixins
        module Channels
          include Ids
          #
          # This method returns a channel ID given a channel name.
          #
          # @option options [channel] :channel
          #   Channel to get ID for, prefixed with #.
          def channels_id(options = {})
            name = options[:channel]
            throw ArgumentError.new('Required arguments :channel missing') if name.nil?

            id_for(:channel, name, '#', :channels, 'channel_not_found') do
              channels_list
            end
          end
        end
      end
    end
  end
end
```

It appears that the use of channels_list, which is deprecated, causes this failure.

wasabigeek:
Fixes #328 

TBH I'm not sure if I missed something. The two methods are not like-for-like but `conversations_list` seems a reasonable drop-in:
- the default will only return public_channels
- the objects returned seem to have a similar structure to `channels_list`, I would assume the default would filter out all the non-channel objects (e.g. IMs)

For comparison:
- https://api.slack.com/methods/channels.list
- https://api.slack.com/methods/conversations.list


dblock:
I think you're also missing new mixins in https://github.com/slack-ruby/slack-ruby-client/tree/master/lib/slack/web/api/mixins for at least the conversations methods.

lib/slack/web/api/mixins/channels.id.rb
```diff
@@ -17,7 +17,7 @@ def channels_id(options = {})
             throw ArgumentError.new('Required arguments :channel missing') if name.nil?
 
             id_for(:channel, name, '#', :channels, 'channel_not_found') do
-              channels_list
+              conversations_list
```
dblock, line 20: When users call `channels_` they get deprecations. I think we want to keep that. For those that call `conversations_` we shouldn't have any deprecations. I don't think we should mix both. It's in `id_for` that we want to call the new(er) methods.

wasabigeek: I didn't quite follow this
> It's in id_for that we want to call the new(er) methods

If I understood correctly, the `conversations_`-methods still use this same `channels_id` method, which would result in calling the deprecated `channels_list` method. Did you mean we should have some logic to determine whether `channels_list` or `conversations_list` is used, depending on whether it's a `channels_` or `conversations_` method call?

If so, I could make a copy of this method that uses `conversations_list`, and in the method templates use `deprecated: true` as a signal for whether to use this new method instead of `channels_id`. What do you think?

dblock: I'll try again.

- introduce `Mixins::Conversations::Ids#conversations_id` that looks just like this method except change channels to conversations
- deprecate `channels_id` the usual way with a warning
- call `conversations_id` everywhere we call `channels_id` except in deprecated `channels_list` method

This way we don't change this mix-in and fix any method that has not been deprecated.

wasabigeek: Sounds good, will go with that!

wasabigeek:
@dblock there's a big set of changes that are incoming with the updated `slack-api-ref`, could you take a look at this before I push them? I will try to re-record the VCR for `channels_info.yml` after.

wasabigeek:
Hmm on second look the overall impact of this was the same as the previous implementation (swapping out `channels_list` for `conversations_list`) as `channels_list` itself doesn’t use `channels_id` (well, duh :facepalm:), so the Channels mixin is essentially unused now. Should I revert to the previous implementation?

dblock:
As is this looks good to me (other than the failing build, rubocop), what am I missing? Channels mixin can be left alone, the template change makes sense as it defaults to the newer method now.

lib/slack/web/api/templates/method.erb
```diff
@@ -40,7 +40,7 @@ module Slack
 <% if data['group'] == 'groups' && data['args']['channel'] && !data['args']['channel']['desc'].include?('Can be an encoded ID, or a name.') %>
             options = options.merge(channel: groups_id(options)['group']['id']) if options[:channel]
 <% elsif data['args']['channel'] && !data['args']['channel']['desc'].include?('Can be an encoded ID, or a name.') %>
-            options = options.merge(channel: channels_id(options)['channel']['id']) if options[:channel]
+            options = options.merge(channel: conversations_id(options)['channel']['id']) if options[:channel]
```
dblock, line 43: This is perfect, fixes the regression in all cases.

wasabigeek:
> As is this looks good to me (other than the failing build, rubocop), what am I missing? Channels mixin can be left alone, the template change makes sense as it defaults to the newer method now.

It seemed like the Channels mixin becomes essentially useless, as it's one and only method (`channels_id`) is no longer called elsewhere in the code (except in bin/commands/channels.rb, I'm not sure what that is for yet), so I thought we may as well get rid of it / repurpose it to use `conversations_list` - what do you think?

dblock:
> > As is this looks good to me (other than the failing build, rubocop), what am I missing? Channels mixin can be left alone, the template change makes sense as it defaults to the newer method now.
> 
> It seemed like the Channels mixin becomes essentially useless, as it's one and only method (`channels_id`) is no longer called elsewhere in the code (except in bin/commands/channels.rb, I'm not sure what that is for yet), so I thought we may as well get rid of it / repurpose it to use `conversations_list` - what do you think?

I think someone may be using it directly so we should keep it. We can delete it when the deprecated API is removed I think.

Happy to merge on 🍏

spec/slack/web/api/endpoints/custom_specs/channels_spec.rb
```diff
@@ -5,9 +5,9 @@
   let(:client) { Slack::Web::Client.new }
 
   context 'channels' do
-    it 'info', vcr: { cassette_name: 'web/channels_info' } do
-      json = client.channels_info(channel: '#general')
-      expect(json.channel.name).to eq 'general'
+    it 'raises deprecation error', vcr: { cassette_name: 'web/channels_info' } do
+      expect { client.channels_info(channel: '#general') }
+        .to raise_error(Slack::Web::Api::Errors::MethodDeprecated, 'method_deprecated')
```
wasabigeek, line 10: Note the new recording now raises a deprecated error for the underlying `channels_info` method itself, I don't have a legacy token to try recording what it would look like for that :/

spec/fixtures/slack/web/channels_info.yml
```diff
@@ -2,45 +2,138 @@
...
       string: token=token
+    headers:
+      Accept:
+      - application/json; charset=utf-8
+      User-Agent:
+      - Slack Ruby Client/0.15.0
+      Content-Type:
+      - application/x-www-form-urlencoded
+      Accept-Encoding:
+      - gzip;q=1.0,deflate;q=0.6,identity;q=0.3
```
wasabigeek, line 17: Do you think it's useful to leave all these additional headers in?

dblock: I don't think it matters. I rarely edit anything in VCR.

spec/slack/web/api/endpoints/custom_specs/channels_spec.rb
dblock: It's fine

dblock:
Not sure what's up with CI.

dblock:
Amen, merged.

PR #336, opened 2020-08-02
don't allways fill events.

pyama86:
Hi.
I notice this log which from `slack-ruby-bot` output.
```
#run_handlers: undefined method `[]' for nil:NilClass (NoMethodError)
away-from-keyboard-59cb64c5db-n76g7 away-from-keyboard-bot /usr/local/bundle/ruby/2.6.0/gems/slack-ruby-client-0.15.0/lib/slack/real_time/client.rb:254:in `run_handlers'
away-from-keyboard-59cb64c5db-n76g7 away-from-keyboard-bot /usr/local/bundle/ruby/2.6.0/gems/slack-ruby-client-0.15.0/lib/slack/real_time/client.rb:246:in `dispatch'
```
I don't always defined any hooks, so I add nil check before run_handler.

thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ### 0.15.1 (Next)
 
+* Your contribution here.
 * [#336](https://github.com/slack-ruby/slack-ruby-client/pull/336): Don't allways fill events - [@pyama86](https://github.com/pyama86).
 
 ### 0.15.0 (2020/7/26)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ### 0.15.1 (Next)
 
 * Your contribution here.
-* [#336](https://github.com/slack-ruby/slack-ruby-client/pull/336): Don't allways fill events - [@pyama86](https://github.com/pyama86).
+* [#336](https://github.com/slack-ruby/slack-ruby-client/pull/336): Fix: handle nil events - [@pyama86](https://github.com/pyama86).
 
 ### 0.15.0 (2020/7/26)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ### 0.15.1 (Next)
 
-* Your contribution here.
+* [#336](https://github.com/slack-ruby/slack-ruby-client/pull/336): Don't allways fill events - [@pyama86](https://github.com/pyama86).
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ### 0.15.1 (Next)
 
-* Your contribution here.
```
dblock, line 3: danger wants you to put this back

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
...
+* [#336](https://github.com/slack-ruby/slack-ruby-client/pull/336): Don't allways fill events - [@pyama86](https://github.com/pyama86).
```
dblock, line 3: Maybe "Fix: handle `nil` events"?

lib/slack/real_time/client.rb
```diff
@@ -251,9 +251,11 @@ def dispatch(event)
       end
 
       def run_handlers(type, data)
-        handlers = store.class.events[type.to_s]
-        handlers&.each do |handler|
-          store.instance_exec(data, &handler)
+        if store.class.events
```
dblock, line 254:
```ruby
return unless store.class.events
...
```

lib/slack/real_time/client.rb
```diff
@@ -251,9 +251,11 @@ def dispatch(event)
...
+          handlers = store.class.events[type.to_s]
+          handlers&.each do |handler|
```
dblock, line 256:
```ruby
return unless handlers
```

... to be consistent

lib/slack/real_time/client.rb
```diff
@@ -251,7 +251,11 @@ def dispatch(event)
       end
 
       def run_handlers(type, data)
+        return unless store.class.events
+
         handlers = store.class.events[type.to_s]
+        return unless handlers
+
         handlers&.each do |handler|
```
dblock, line 259: No longer need `&` here.

pyama86:
> didn't look at the detail.
It seems that the test failed in a place that has nothing to do with my change.

dblock:
> > didn't look at the detail.
> 
> It seems that the test failed in a place that has nothing to do with my change.

We still need to get the build (back to) green first. I will take a look when I have time, or you can :)

dblock:
Now it all makes sense, merged ;)

PR #339, opened 2020-08-24
Fix #337: channel_not_found error resolving channel IDs with 100+ channels

### LINKED ISSUE #337: Channel_not_found in new conversations_list usage
## Problem
On the latest release of 0.15.0 (I upgraded due to the deprecated channels_list to use the new conversations_list) I'm seeing an error when I try to query by a channels name rather than a channel_id. I expect to be able to just `client.conversations_setTopic({:channel => '#channel_name', :topic => 'random topic})` <- but the hashtag channel name format no longer works and only the channel_id approach works.

I find that the code is much more readable with channel names than ids, so please help!

image

dblock:
I made myself some channels and was able to reproduce this.


dblock:
@wasabigeek care to CR?

wasabigeek (approved):

wasabigeek:
LGTM, this is a lot easier than what I was doing >_<

CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 ### 0.15.1 (Next)
 
-* Your contribution here.
 * [#336](https://github.com/slack-ruby/slack-ruby-client/pull/336): Fix: handle nil events - [@pyama86](https://github.com/pyama86).
+* [#339](https://github.com/slack-ruby/slack-ruby-client/pull/339): Fix: `channel_not_found` resolving channel IDs with 100+ channels - [@dblock](https://github.com/dblock).
```
wasabigeek, line 4: Potentially with all paginated list endpoints as well

dblock: Right. I'll just leave it like this.

PR #340, opened 2020-09-01
Allow setting the faraday adapter for connections

watsonjon:
Most of the time using the default Faraday adapter is fine, but _sometimes_ we want to allow that to differ. This will allow that to be set as part of the client configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,7 @@ timeout             | Optional open/read timeout in seconds.
 open_timeout        | Optional connection open timeout in seconds.
 default_page_size   | Optional page size for paginated requests, default is _100_.
 default_max_retries | Optional number of retries for paginated requests, default is _100_.
+adapter             | Optional HTTP adapter to use, defaults to Faraday.default_adapter.
 
 You can also pass request options, including `timeout` and `open_timeout` into individual calls.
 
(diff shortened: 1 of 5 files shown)


dblock:
LGTM, needs README and CHANGELOG please

README.md
```diff
@@ -240,6 +240,7 @@ timeout             | Optional open/read timeout in seconds.
 open_timeout        | Optional connection open timeout in seconds.
 default_page_size   | Optional page size for paginated requests, default is _100_.
 default_max_retries | Optional number of retries for paginated requests, default is _100_.
+adapter             | Optional HTTP adapter to use, defaults to Faraday.default_adapter.
```
dblock, line 243: Make this either underline, or `code` like the other defaults, please.

dblock:
I'm curious why one would want to change the client's default adapter and not `Faraday.default_adapter = ...`?

spec/slack/web/client_spec.rb
```diff
@@ -150,6 +150,47 @@
       end
     end
 
+    context 'adapter option' do
+      let(:adapter) { Faraday.default_adapter }
+      let(:adapter_class) { Faraday::Adapter::NetHttp }
+
+      before do
```
dblock, line 157: You need to reset the default adapter to the old value around the spec to avoid polluting other spec results.

watsonjon:
@dblock we use `:typhoeus` as the default adapter for most of our `Faraday` connections but there are cases where we need a connection to use `:net_http` so it is not using curl due to HTTP2 issues that crop up from time to time. We run into that with our slack connections where we get `Error in the HTTP2 framing layer Failed to send slack notification`. We can monkey patch it but seemed better to allow others to configure that as well.

dblock:
Merged, thank you.

Issue #321, opened 2020-05-04
undefined method `new' for BigDecimal:Class, Ruby 2.7


geraldeersteling:
Hello,

I'm trying to use the client in combination with fastlane, this worked fine up till today.

When I try to `require` the gem it immediately crashes with the following exception:

`NoMethodError: [!] undefined method `new' for BigDecimal:Class`

The complete stack trace is as follows:

```
bundler: failed to load command: fastlane (/Users/REDACTED/.gem/ruby/2.7.0/bin/fastlane)
NoMethodError: [!] undefined method `new' for BigDecimal:Class
  /Users/REDACTED/.gem/ruby/2.7.0/gems/activesupport-4.2.11.1/lib/active_support/core_ext/object/duplicable.rb:111:in `<class:BigDecimal>'
  /Users/REDACTED/.gem/ruby/2.7.0/gems/activesupport-4.2.11.1/lib/active_support/core_ext/object/duplicable.rb:106:in `<top (required)>'
  /Users/REDACTED/.gem/ruby/2.7.0/gems/activesupport-4.2.11.1/lib/active_support/core_ext/object.rb:3:in `require'
  /Users/REDACTED/.gem/ruby/2.7.0/gems/activesupport-4.2.11.1/lib/active_support/core_ext/object.rb:3:in `<top (required)>'
  /Users/REDACTED/.gem/ruby/2.7.0/gems/activesupport-4.2.11.1/lib/active_support/core_ext.rb:2:in `require'
  /Users/REDACTED/.gem/ruby/2.7.0/gems/activesupport-4.2.11.1/lib/active_support/core_ext.rb:2:in `block in <top (required)>'
  /Users/REDACTED/.gem/ruby/2.7.0/gems/activesupport-4.2.11.1/lib/active_support/core_ext.rb:1:in `each'
  /Users/REDACTED/.gem/ruby/2.7.0/gems/activesupport-4.2.11.1/lib/active_support/core_ext.rb:1:in `<top (required)>'
  /Users/REDACTED/.gem/ruby/2.7.0/gems/slack-ruby-client-0.14.4/lib/slack-ruby-client.rb:39:in `require'
  /Users/REDACTED/.gem/ruby/2.7.0/gems/slack-ruby-client-0.14.4/lib/slack-ruby-client.rb:39:in `<top (required)>'
[...]

PR #344, opened 2020-09-04
Add ruby-2.7.1 to CI

wasabigeek:
Super early look at https://github.com/slack-ruby/slack-ruby-client/issues/321

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ cache: bundler
 rvm:
   - 2.5.3
   - 2.6.0
+  - 2.7.1
   - ruby-head
   - jruby-head
 


wasabigeek:
@dblock should I also add a test run with a concurrency gem?

dblock:
> @dblock should I also add a test run with a concurrency gem?

I know nothing about this. Qu'est-ce que c'est?

wasabigeek:
Oh I meant something like this:
```
    - rvm: 2.7.1
      env: CONCURRENCY=async-websocket
```

dblock:
> Oh I meant something like this:
> 
> ```
>     - rvm: 2.7.1
>       env: CONCURRENCY=async-websocket
> ```

Sure why not. But let's not explode the matrix.

wasabigeek:
Maybe I'll leave it like this? If I'm understanding correctly the non-matrix runs are for testing against ruby versions, the matrix runs are for danger / with concurrency gem

dblock:
We can start with this, merged.

PR #349, opened 2020-10-07
Safely include quoted examples.

dblock:
@wasabigeek i think this is cleaner, wdyt?

### DIFF
--- a/lib/slack/web/api/templates/method_spec.erb
+++ b/lib/slack/web/api/templates/method_spec.erb
@@ -12,7 +12,7 @@ RSpec.describe Slack::Web::Api::Endpoints::<%= group.gsub(".", "_").camelize %> 
   context '<%= group %>_<%= name %>' do
 <% required_params.each do |arg_name, arg_v| %>
     it 'requires <%= arg_name %>' do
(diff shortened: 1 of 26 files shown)


wasabigeek (approved):
😲 didn't know you could do that!

lib/slack/web/api/templates/method_spec.erb
```diff
@@ -12,7 +12,7 @@ RSpec.describe Slack::Web::Api::Endpoints::<%= group.gsub(".", "_").camelize %>
   context '<%= group %>_<%= name %>' do
 <% required_params.each do |arg_name, arg_v| %>
     it 'requires <%= arg_name %>' do
-<% params_except_required = required_params.reject{ |name, _| name == arg_name  }.map{|var, opts| "#{var}: '#{opts['example'].gsub("'", '"')}'"}.join(', ') %>
+<% params_except_required = required_params.reject{ |name, _| name == arg_name  }.map{ |var, opts| "#{var}: %q[#{opts['example']}]"}.join(', ') %>
```
wasabigeek, line 15: Hmm the docs seem to suggest that interpolation needs the capital %Q, does this work?

dblock: But we don't want to interpolate here, do we? That string is a copy-paste of an example.

wasabigeek: Ah ok my bad - brain having some trouble reconciling that the %q is part of the template >_<

Issue #306, opened 2019-12-06
Slack 500 page error handling


jcheatham:
Hi!  In the process of interacting with Slack's API, it sometimes has the occasion to return an html 500 status page which then causes a `Faraday::ParsingError` with something like
```
unexpected token at '<!DOCTYPE html> <html lang="en"> <head> <meta charset="utf-8"> <title>Server Error | Slack</title> <meta name="author" content="Slack"> etc.
```

Since the gem is speccing out its own error namespace, it feels like it'd be more consistent to handle and re-raise something in the `Slack::Web::Api::Error` domain instead.  Thanks!

dblock:
Yes, indeed. On a 500 we should be reporting something other than a parsing error. Please contribute?

Deconstrained:
@dblock I've been looking into where in the stack/middleware to implement a rescue/raise for this, and so far the most straightforward solution is in [Slack::Web::Faraday::Request.request](https://github.com/slack-ruby/slack-ruby-client/blob/master/lib/slack/web/faraday/request.rb#L24-L37).

Do you agree with this approach?

dblock:
Usually these things are handled by middleware, eg. https://github.com/slack-ruby/slack-ruby-client/blob/master/lib/slack/web/faraday/response/raise_error.rb

Start by writing a test and solving the problem in any way you can. We can improve upon it.

Deconstrained:
I had been thinking along those lines as well, but it has been challenging to grok the order of middleware execution.

So, in this PR I built a test and explored a non-`rescue` method of detecting an invalid response and raising. Please let me know what you think:

https://github.com/slack-ruby/slack-ruby-client/pull/307

dblock:
Implemented in https://github.com/slack-ruby/slack-ruby-client/pull/350

PR #350, opened 2020-10-12
Handle server errors such as timouts & non-json responses

ojab:
My stab @ https://github.com/slack-ruby/slack-ruby-client/issues/306
It's really a draft, but I want some feedback since I'm not really used to slack-ruby-client, so making it non-draft.


lib/slack/web/faraday/response/raise_error.rb
```diff
@@ -15,6 +15,14 @@ def on_complete(env)
             error_class ||= Slack::Web::Api::Errors::SlackError
             raise error_class.new(error_message, env.response)
           end
+
+          def call(env)
+            super
+          rescue Slack::Web::Api::Errors::SlackError, Slack::Web::Api::Errors::TooManyRequestsError => e
```
dblock, line 21: I think `TooManyRequestsError` is an `SlackError` so both here would get rescued with just `SlackError`

lib/slack/web/faraday/response/raise_error.rb
```diff
@@ -15,6 +15,14 @@ def on_complete(env)
...
+            raise e
+          rescue ::Faraday::ServerError, ::Faraday::ParsingError
+            raise Slack::Web::Api::Errors::InternalError.new('internal_error', env.response)
```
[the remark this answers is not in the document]
dblock: I think you want to do the inverse here, handle only errors that are predictable. We don't want to rescue all faraday errors and narrow them down into an `InternalError`, but we do want to rescue predictable situations, such as the server had an internal error and spit out a bunch of non-JSON into a specific `Slack::Web::Api::Errors::InternalError`. The latter should have an inner exception that gets assigned `e`, along with `env.response`, and the message should be one from `::Faraday::ParsingError` I imagine.

dblock (changes requested):

lib/slack/web/faraday/response/raise_error.rb
ojab:
```
> Slack::Web::Api::Errors::TooManyRequestsError <  Slack::Web::Api::Errors::SlackError
=> nil
```

dblock: Ah yes because we inherited this one from `Faraday::Error` to get `retry_after`.

dblock: Maybe there needs to be an intermediate `Slack::Web::Api::Errors::HttpRequestError` that holds an inner `FaradayError` exception?

lib/slack/web/faraday/response/raise_error.rb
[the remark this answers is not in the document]
ojab: From the testcase in spec:
```
[2] pry(#<RSpec::ExampleGroups::SlackWebFaradayResponseRaiseError::Call::WithNonJsonReply>)> _ex_
=> #<Slack::Web::Api::Errors::InternalError #<Slack::Web::Api::Errors::InternalError: internal_error>>
[3] pry(#<RSpec::ExampleGroups::SlackWebFaradayResponseRaiseError::Call::WithNonJsonReply>)> _ex_.cause
=> #<Faraday::ParsingError #<Faraday::ParsingError: oops>>
```
So we have Faraday error in `e.cause` and can do anything with it anyway. 

Could you please elaborate on `We don't want to rescue all faraday errors and narrow them down into an InternalError, but we do want to rescue predictable situations, such as the server had an internal error and spit out a bunch of non-JSON into a specific Slack::Web::Api::Errors::InternalError.`?
Should we subclass `Slack::Web::Api::Errors::InternalError` into multiple ones?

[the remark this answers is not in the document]
dblock: You're rescuing any faraday error that has specifics to it (e.g. timed out, parsing failed, etc.) into some generic internal error. That seems like it's providing less value than before (narrows it down). Unless I'm misunderstanding something.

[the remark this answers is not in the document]
ojab: Well, we cannot predict what errors slack will fails with, but `::Faraday::ServerError` is basically 5xx & timeouts (which we should handle) and `::Faraday::ParsingError` is non-json (which we also should handle). So I don't think that we.re overrescuing here.

Faraday exception is always available via `e.cause`, source exception (i. e. `Net::ReadTimeout`) is available via `e.cause.cause`, thus even while we're narrowing info in the exception itself, no info is lost and everything is accessible if needed.

We can do something like (naming i hard)
```ruby
module Slack::Web::Api::Errors
  class ServerError < InternalError; end # common class
  class ParsingError < Server; end # wrap ::Faraday::ParsingError
  class ConnectionError < InternalError; end # just commong class
  class TimeoutError < ConnectionError; end # For ::Faraday::ServerError where cause is timeout
  class UnavailableError < ConnectionError; end # For ::Faraday::ServerError where status is 5xx
end
```
if you wish & raise different exceptions for different causes, but still be able to catch them via `rescue Slack::Web::Api::Errors::Error` or more precise ones.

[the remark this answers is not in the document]
dblock: I like this, but I am not sure what the right answer is here, put up a PR that you think should/is ready to be merged and I'll give it more thought?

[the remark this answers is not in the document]
ojab: Pushed, not sure about naming/messages in errors and there is behaviour change: we're raising `SlackError` instead of `Faraday::Error` if server is unavailable, I can revert that one if you wish (and fix the typo in the disabled spec).

lib/slack/web/faraday/response/raise_error.rb
```diff
@@ -6,7 +6,7 @@ module Response
         class RaiseError < ::Faraday::Response::Middleware
           def on_complete(env)
             raise Slack::Web::Api::Errors::TooManyRequestsError, env.response if env.status == 429
-            return unless (body = env.body) && !body['ok']
+            return if (body = env.body).nil? || body['ok'] || !env.success?
```
dblock, line 9: Are we swallowing errors here when `!env.success?`? Or does that get raised as a Faraday error?

I would break this up for readability:

```ruby
return unless env.success?

body = env.body
return unless body
return if body['ok']
```

spec/slack/web/api/errors/service_unavailable_spec.rb
```diff
@@ -4,8 +4,8 @@
 RSpec.describe Slack::Web::Client do
   let(:client) { described_class.new }
 
-  it 'raises a Faraday::ClientError when Slack is unavailable',
-     vcr: { cassette_name: 'web/503_error' } do
+  xit 'raises a Faraday::ClientError when Slack is unavailable',
```
dblock, line 7: Leftover

lib/slack/web/faraday/response/raise_error.rb
```diff
@@ -15,6 +15,16 @@ def on_complete(env)
             error_class ||= Slack::Web::Api::Errors::SlackError
             raise error_class.new(error_message, env.response)
           end
+
+          def call(env)
+            super
+          rescue Slack::Web::Api::Errors::SlackError, Slack::Web::Api::Errors::TooManyRequestsError
+            raise
+          rescue ::Faraday::ParsingError
+            raise Slack::Web::Api::Errors::ParsingError.new('parsing_error', env.response)
```
dblock, line 24: I would sink `parsing_error` message into the error definition itself when calling `super`

dblock (changes requested):
Needs a clean green build please, README and UPGRADING updates. Otherwise LGTM.

lib/slack/web/faraday/response/raise_error.rb
ojab: Not sure what do you mean here.

.rubocop.yml
```diff
@@ -19,6 +19,8 @@ Layout/EmptyLineAfterMagicComment:
 
 Metrics/BlockLength:
   Max: 250
+  Exclude:
```
dblock, line 22: We can just `Enable: false` for this one.

lib/slack/web/faraday/response/raise_error.rb
dblock: Like this. I don't think there's some kind of convention that the first argument must be a string, but I could be wrong? This is not a must have change for me here.

```ruby
raise Slack::Web::Api::Errors::ParsingError.new(env.response)
```

UPGRADING.md
```diff
@@ -1,6 +1,12 @@
 Upgrading Slack-Ruby-Client
 ===========================
 
+### Upgrading to >= 0.16.0
+
+As of 0.16.0 `Faraday::Error` exceptions sans `Faraday::ClientError` are wrapped into `Slack::Web::Api::Errors::ServerError`.
+
+See [README#other-errors](README.md#other-errors) and [#350](https://github.com/slack-ruby/slack-ruby-client/pull/350) for more information.
```
dblock, line 8: We need to tell people that were rescuing `Faraday::Error` or similar how their code needs to change.

dblock (changes requested):
See my comments, expand on UPGRADING, add a CHANGELOG entry and this is good to go. Thanks for hanging in here!

lib/slack/web/faraday/response/raise_error.rb
ojab: As a user of this gem I'm expecting first argument (i. e. message) of `Slack::Web::Api::Errors::Error` to be a kinda meaningful `String`, while the second is response, that's how it works everywhere currently. So I assume that there actually is an implicit convention.
And while it will be more appropriate to catch & process `Slack::Web::Api::Errors::ParsingError` separately — I assume something like
```ruby
rescue Slack::Web::Api::Errors::Error => e
  raise unless e.message == 'parsing_error'
```
also is expected to work.

But I also don't have strong feelings about this.

dblock: My issue is that it's an argument to the constructor that hides the message argument. I was thinking like so:

```ruby
class TimeoutError < HttpRequestError
  def initialize(response)
     super 'timeout_error'
  end
end
```

```ruby
raise Slack::Web::Api::Errors::TimeoutError
```

dblock:
Needs a CHANGELOG entry otherwise good to go with or without the message/super change.

ojab:
CHANGELOG is added.

dblock:
Merged

Issue #186, opened 2017-11-29
warning: method redefined; discarding old concurrency


aadityataparia:
`/lib/ruby/gems/2.4.0/gems/slack-ruby-client-0.10.0/lib/slack/real_time/config.rb:32: warning: method redefined; discarding old concurrency`
Seems like method concurrency has been defined at two places!

dblock:
That looks like a bug, without any bad side effects but still, would appreciate a fix.

aadityataparia:
One problem I had with this was CI marked test as error, because this warning was shown.

dblock:
@aadityataparia Care to help with a PR?

aadityataparia:
I am not able to find where it is defined before?

dblock:
I am going to reopen this since it looks like a real problem. If you want to fix it @aadityataparia you're more than welcome. Someone else reading this can help too.

PR #353, opened 2020-11-13
Drop redefined accessor methods

danhalligan:
This drops redefined accessor methods:
- `concurrency` is explicitly defined in `lib/slack/real_time/config.rb`
- `logger` in `lib/slack/real_time/client.rb` is defined by the line `attr_accessor(*Config::ATTRIBUTES)` above.

Resolves: #186

### DIFF
--- a/lib/slack/real_time/client.rb
+++ b/lib/slack/real_time/client.rb
@@ -20,7 +20,6 @@ module Slack
       attr_accessor :store
       attr_accessor :url
       attr_accessor(*Config::ATTRIBUTES)
-      attr_accessor :logger
 
       protected :store_class, :store_class=
 
--- a/lib/slack/real_time/config.rb
+++ b/lib/slack/real_time/config.rb
@@ -10,13 +10,14 @@ module Slack
         token
         websocket_ping
         websocket_proxy
+        concurrency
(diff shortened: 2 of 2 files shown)


dblock:
Looks like legit build failures:

```
     Failure/Error: concurrency::Socket
      
      NameError:
        undefined local variable or method `concurrency' for #<Slack::RealTime::Client:0x000000000110e928>
```

danhalligan:
I believe this is fixed now. `Config::ATTRIBUTES` is used elsewhere to assign readers/writers. To fix the warning due to duplicated method definition in `lib/slack/real_time/config.rb`, I just removed `:concurrency` from attributes and then defined just the writer, leaving the reader to be defined later in the file.

dblock:
It's a little ugly, but I'll take it. Thanks.

danhalligan:
> No object is so beautiful that, under certain conditions, it will not look ugly.

dblock:
Well, if you want to clean this up, remove `:concurrency` from `Config::ATTRIBUTES` (move it), then add it where needed.

danhalligan:
That's a fair point. I'll have a look if I get time.

PR #354, opened 2020-12-19
Rewind body after checking request signature

sunny:
## Before

Trying to read the request body after calling a Slack verification, returns an empty body:

```rb
Slack::Events::Request.new(request).verify!
request.body.read # => ""
```

This is because calling `request.body.read` on Rack requests causes the body’s StringIO to go to the end of the string.

## After

Therefore, to minimize surprise, we can rewind the body after reading it. After this fix, we get the following:

```rb
Slack::Events::Request.new(request).verify!
request.body.read # => "payload=%7B%22type%22%3A%22blo…"
```


dblock:
Thanks, makes sense!

Issue #338, opened 2020-08-05
Remove support for multiple concurrency libraries


dblock:
I suspect nobody uses celluloid anymore and we have a broken implementation with faye-websocket, #285, so nobody uses that either. Cut down on maintenance and remove those when upgrading to #282.

dblock:
@wasabigeek want to remove faye the same way? We should still leave concurrency optional because not everybody needs RTM, so the doc should just say that.

PR #357, opened 2021-01-21
Remove faye

wasabigeek:
Fixes https://github.com/slack-ruby/slack-ruby-client/issues/338


dblock:
Rebase, squash, and this is good to go! Thank you.

PR #370, opened 2021-03-06
Remove deprecated methods

### LINKED ISSUE #365: Remove deprecated methods
The `im_open` and `channels_list` methods no longer work since February 2021, purge them from the library.

dblock:
Closes #365


bin/commands/channels.rb
```diff
@@ -53,128 +20,4 @@
...
-  end
-
-  g.desc 'Joins a channel, creating it if needed.'
-  g.long_desc %( Joins a channel, creating it if needed. )
-  g.command 'join' do |c|
-    c.flag 'name', desc: 'Name of channel to join.'
-    c.flag 'team_id', desc: 'encoded team id to list channels in, required if org token is used.'
-    c.flag 'validate', desc: 'Whether to return errors on invalid channel name instead of modifying it to meet the specified criteria.'
-    c.action do |_global_options, options, _args|
-      puts JSON.dump($client.channels_join(options))
```
DariuszMusielak, line 84: I have been using `channels_join`. What else should I use know since this one is depricated?

chrisbloom7: @DariuszMusielak I believe https://api.slack.com/methods/conversations.join is what you want to use (or `conversations_join` in the client). I can't find reference to any previous `channels.join` method in the latest API docs, but other `channels.*` methods have all been deprecated in favor of the `conversations` namespace.

PR #374, opened 2021-05-06
Enable the logging of the request body

ts-3156:
When debugging, I sometimes want to see the request body. I don't know if other people feel the same way. If you don't need it, please feel free to close this pull request.

### DIFF
--- a/lib/slack/web/faraday/connection.rb
+++ b/lib/slack/web/faraday/connection.rb
@@ -28,7 +28,7 @@ module Slack
                 connection.use ::FaradayMiddleware::Mashify, mash_class: Slack::Messages::Message
                 connection.use ::FaradayMiddleware::ParseJson
                 connection.use ::Slack::Web::Faraday::Response::WrapError
-                connection.response :logger, logger if logger
+                connection.response :logger, logger, bodies: true if logger
                 connection.adapter adapter
               end
             end


dblock:
I think we don't want this by default. Did you figure out a way to do this in your own code? I would like us to document how to do this without changing the default configuration, in README.

ts-3156:
@dblock Thank you for taking the time to reply.

> I think we don't want this by default.

I have the same opinion. However, I have not come up with a suitable code to pass a value for `bodies: true` via `Slack.configure`.

```
Slack.configure do |config|
  config.log_bodies = true # <- This name feels strange.
end
```


> Did you figure out a way to do this in your own code?

Yes. I found this option when I investigated strange behavior that Slack doesn't display the image preview correctly.


> I would like us to document how to do this without changing the default configuration, in README.

If you need this option for logging, I will write how to do it in README. However, in order to use this option, I need to modify not just the documentation, but also the code. Is this really what you want?

You don't need to hurry. It is OK if this idea is discarded.

dblock:
How about introducing `logger_options`?

If you feel more ambitious, take a look at https://lostisland.github.io/faraday/middleware/logger. It also supports blocks and filtering which we could expose. The https://github.com/ashkan18/graphlient gem does it like so:

```ruby
 Graphlient::Client.new('http://test-graphql.biz/graphql') do |client|
      client.http do |h|
        h.connection do |c|
          c.adapter Faraday::Adapter::Rack, app
        end
      end
    end
```

So `http` and `connection` yield from within the configuration and you would want to be able to write the following, overriding the `:logger` part of the configuration.

```ruby
 Slack.configure do |config|
   config.connection do |conn|
     c.response :logger, logger, { bodies: true }
   end
 end
end
```

alex-benoit:
This is very interesting to implement. Sometimes I get the error `Slack::Web::Api::Errors::InvalidArguments` but would like to have a look at the exact block that created the error in the `response_metadata`. For some reason, this is not captured automatically by Sentry. I would like to pass the Faraday client response as breadcrumbs.

PR #380, opened 2021-06-19
ServerError does not descend from InternalError

### LINKED ISSUE #373: Suggestion: ServerError should not descend from InternalError or SlackError
**tl;dr below**

I love the addition of `ServerError` and its descendants in #350, and the further refinement in #359. (In fact I'd like to take it further and wrap `Faraday::ClientError` in a new class as well). However it seems wrong to me that `ServerError` descends from `InternalError` and in turn from `SlackError`.

Prior to #350, `SlackError` and its many descendants represented a specific type of error: the Web API has responded with a 200, but the body has `"ok":false` and `"error":"some_error_code"`. In other words, the API has received the request, processed it, and for some reason the API cannot comply. #312 (which I authored) added the specific descendant classes for all known `error` codes, but still `SlackError` only represented 200 with `"ok":false`.

All of these new error classes represent wholly different situations, in some cases a 500, or in other cases a 200 with a non-JSON body (representing something like an outage). From a classification standpoint it would make more sense to give these new errors a dedicated parent class and leave `SlackError` to represent just the `"ok":false` responses.

There's also a practical consideration, which is that it appears the 3 methods defined on `SlackError` will raise if accessed on these new error classes, since `response.body` doesn't conform. (I could be wrong about this, I haven't actually tried it.) Those methods are specific to the body from a response that has `"ok":false`.

**begin tl;dr**
My suggestion is to have `Slack::Web::Api::Errors::ServerError` descend directly from `::Faraday::Error`
 (just like `SlackError` does). This would be a breaking change and require the appropriate version bump.
**end tl;dr**

Here's another advantage to this change: I saw @dblock's [comment on #350](https://github.com/slack-ruby/slack-ruby-client/pull/350/files#r509196923) about the redundancy of passing `parsing_error` into `new`, which is necessary for the initializer on `SlackError`. With this new approach `ServerError` (and perhaps the descendants) could define its own initializer that doesn't require this.

One final note: this pattern is already established by `TooManyRequestsError` which descends directly from `::Faraday::Error` rather than `SlackError`. Like the situations covered by `ServerError` it is not a 200 response that has `"ok":false`, which I assume is why it was made to not extend `SlackError`.

Tagging also @ojab and @agrobbin as the authors of #350 and #359, respectively.

jmanian:
Resolves #373.

Notes:
1. The change in class hierarchy is a breaking change for anyone who might be rescuing `SlackError` and expecting it to rescue `ServerError`.
2. I took the [suggestion](https://github.com/slack-ruby/slack-ruby-client/pull/350/files#r509196923) from #350 about syncing the message into the class definition rather than passing it in. This is a breaking change for anyone instantiating these errors in their code.
3. I tweaked `TooManyRequestsError` to follow the same pattern for defining its `message`. Not a breaking change, just implementation.

Questions:
1. Is it OK for these 5 error classes to share a file now that they're no longer empty definitions?
2. I didn't do it here, but I'm tempted to change these 3 error messages to something more human readable rather than an underscore code. I assume the underscore messages were inspired by the error messages that Slack returns, but since these are not error messages from Slack they don't _need_ to conform, and in fact it could be misleading if they look like they might be messages from Slack. For example `UnavailableError` could be `"The server responded with #{status}"`. Thoughts on this? It could be considered a breaking change.
3. I didn't do it yet, but I'm tempted to follow the same `message` pattern for all the descendants of `SlackError` defined in `errors.rb`, so that instead of `Slack::Web::Api::Errors::AccountInactive.new('account_inactive', response)` it's simply `Slack::Web::Api::Errors::AccountInactive.new(response)`. This would be a breaking change for anyone instantiating these error classes.

To be done later, if this approach is good:
- [x] Add to changelog
- [x] Add an entry to upgrading.md
- [x] Move the version to 0.18.0 since this is a breaking change


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
-### 0.17.1 (Next)
+### 0.18.0 (Next)
 
 * Your contribution here.
+* [#380](https://github.com/slack-ruby/slack-ruby-client/pull/380): Updates to server error classes and hierarchy (see [UPGRADING](UPGRADING.md) for more) - [@jmanian](https://github.com/jmanian).
```
dblock, line 4: No need to mention UPGRADING here, or our CHANGELOG will be littered with them.

README.md
```diff
@@ -324,9 +324,9 @@ If you exceed [Slack’s rate limits](https://api.slack.com/docs/rate-limits), a
 
 ##### Other Errors
 
-In case of Slack temporarily unavailability a `Slack::Web::Api::Errors::ServerError` (`Slack::Web::Api::Errors::SlackError` subclass) subclasses will be raised and original `Faraday::Error` will be accesible via `exception.cause`.
+In case of Slack temporarily unavailability a subclass of `Slack::Web::Api::Errors::ServerError` will be raised and original `Faraday::Error` will be accesible via `exception.cause`. (This is no longer a subclass of `Slack::Web::Api::Errors::SlackError`, see [UPGRADING](UPGRADING.md).)
```
dblock, line 327: Let's write "Since version X.Y.Z this class is no longer ... and remove see UPGRADING.

UPGRADING.md
```diff
@@ -1,6 +1,14 @@
 Upgrading Slack-Ruby-Client
 ===========================
 
+### Upgrading to >= 0.18.0
+
+#### Error Handling
+
+As of 0.18.0 `Slack::Web::Api::Errors::ServerError` and its subclasses (introduced in 0.16.0) no longer extend `Slack::Web::Api::Errors::InternalError` or its parent `Slack::Web::Api::Errors::SlackError`. If you are rescuing `SlackError` or `InternalError` with the intention of including `ServerError` and its subclasses you should adjust your code to explicitly rescue `Slack::Web::Api::Errors::ServerError`.
+
+Additionally the `initialize` method for `ParsingError`, `TimeoutError`, and `UnavailableError` have changed from `new(message, response)` to `new(response)`. The `message` is now built into the definition of these classes. If you are instantiating or raising these errors in your code (perhaps in tests) you will need to update your code.
```
dblock, line 10: Maybe add some before/after?

dblock (approved):
This looks good. Care to cleanup my minor nits and I'll merge? Thank you.

UPGRADING.md
jmanian: You mean example code before and after? I can do that

jmanian:
@dblock Where's Travis?

UPGRADING.md
dblock: Yeah

dblock:
> @dblock Where's Travis?

Not sure. Did they turn off travis-ci.org finally? Want to move us to GitHub actions?

dblock:
@jmanian I enabled travis-ci.com on the org, maybe next push will "just work"? Let's upgrade badges...

jmanian:
@dblock Still no Travis. I originally opened this PR as a draft, maybe that confused it? I authed into Travis to see if I could do anything but the repo doesn't appear for me there.

dblock:
@jmanian I am seeing https://travis-ci.com/github/slack-ruby/slack-ruby-client and it's enabled for PRs. Try making a new PR? Maybe that will kick it? I dunno. Rewrite everything in GitHub actions? :)

jmanian:
@dblock Success!

dblock:
Merged, merci.

PR #395, opened 2021-12-29
Rub CI on Ruby 3.1.0 && update minor versions

ohbarye:
This pull request makes CI runs on Ruby 3.1.0 since Ruby 3.1.0 has been released. Besides, it updates also lower versions.

ref: https://www.ruby-lang.org/en/downloads/releases/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,11 @@ language: ruby
 cache: bundler
 
 rvm:
-  - 2.5.3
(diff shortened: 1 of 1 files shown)


ohbarye:
Ah, I didn't notice that Travis CI hasn't prepared Ruby 3.1.0 yet. https://rubies.travis-ci.org/
This pull request is a little bit early, so I'm closing.

Issue #399, opened 2022-02-23
Dependency on faraday_middleware prevents update to faraday 2.0


brandonrdn:
`faraday_middleware` has been deprecated, with `json` and `instrumentation` middleware being bundled in `faraday 2.0`, and other middleware are being re-released as independent gems. Since `faraday_middleware` is never being updated to support `faraday >= 2.0`, the current `slack-ruby-client` can never use it either. Is there any chance we can remove the `faraday_middleware` dependency and bump the minimum version of `faraday` to `2.0`?

schinery:
I tried to have a go at this myself but the chain of gems that actually needs updating is pretty troublesome...

* This gem needs [slack-ruby-danger](https://github.com/slack-ruby/danger) updated to support Faraday 2.0
* slack-ruby-danger needs [danger](https://github.com/danger/danger) updated to support Faraday 2.0
* danger needs [this issue](https://github.com/danger/danger/issues/1349) closed and a release made to support Faraday 2.0

So, am wondering is it more important to be able to use danger in this repo than users to be able to use Faraday 2.0, which as @pboling [commented](https://github.com/danger/danger/issues/1349#issuecomment-1043243683) "_is far more consequential for the users of oauth2, so it takes precedence_", at least until danger does support Faraday 2.0?

dblock:
The easiest workaround is to move Danger into a separate danger/Gemfile and separate the task in GHA.

schinery:
Ignoring danger for the moment, this is as far as I've got locally trying to get Faraday updated.

**Code changes:**

I've commented out the code that needs replacing rather than removing for the moment. Also ignore the `debugger`, was trying to figure out why the parse error wasn't being thrown. 

```sh
diff --git i/Gemfile w/Gemfile
index 4faec06..adf42a9 100644
--- i/Gemfile
+++ w/Gemfile
@@ -14,14 +14,16 @@ end
 
 group :test do
   gem 'activesupport'
+  gem 'byebug'
   gem 'erubis'
+  gem 'faraday-typhoeus'
   gem 'json-schema'
   gem 'rake', '~> 13'
   gem 'rspec'
   gem 'rubocop', '~> 1.26.0'
   gem 'rubocop-performance'
   gem 'rubocop-rspec'
-  gem 'slack-ruby-danger', '~> 0.2.0', require: false
+  # gem 'slack-ruby-danger', github: "schinery/danger" # '~> 0.2.0', require: false
   gem 'timecop'
[...]

PR #406, opened 2022-05-03
Update Faraday to >= 2.0

schinery:
First attempt at updating Faraday to use `>= 2.0`, which for the most part was fairly straightforward with the following exceptions:

**1: Updating `lib/slack/web/faraday/response/raise_error.rb`**

The spec in `spec/slack/web/client_spec.rb:281` was failing with the following error:

```sh
1) Slack::Web::Client global config server failures parsing error raises ParsingError
     Failure/Error: expect { request }.to raise_error(Slack::Web::Api::Errors::ParsingError).with_message('parsing_error')

       expected Slack::Web::Api::Errors::ParsingError with "parsing_error", got #<NoMethodError: undefined method `map' for nil:NilClass

                                                    ^^^^> with backtrace:
         # ./lib/slack/web/faraday/response/raise_error.rb:18:in `on_complete'
         # ./lib/slack/web/faraday/response/raise_error.rb:26:in `call'
         # ./lib/slack/web/faraday/request.rb:25:in `request'
         # ./lib/slack/web/faraday/request.rb:11:in `post'
         # ./lib/slack/web/api/endpoints/api.rb:17:in `api_test'
         # ./spec/slack/web/client_spec.rb:269:in `block (4 levels) in <top (required)>'
         # ./spec/slack/web/client_spec.rb:282:in `block (6 levels) in <top (required)>'
         # ./spec/slack/web/client_spec.rb:282:in `block (5 levels) in <top (required)>'
     # ./spec/slack/web/client_spec.rb:282:in `block (5 levels) in <top (required)>'

Finished in 10.54 seconds (files took 1.93 seconds to load)
461 examples, 1 failure, 8 pending

Failed examples:

rspec ./spec/slack/web/client_spec.rb:281 # Slack::Web::Client global config server failures parsing error raises ParsingError
```

I believe this is caused by there being no middleware equivalent to `::FaradayMiddleware::ParseJson`, and the new `:json` request/response middleware appears to only parse (and throw `::Faraday::ParsingError`) if the content type is `application/json`.

So, I've added some code to check if the `env.body` has been converted into `Hash` or not, and attempt to convert if it isn't. 

It does raise the question does `::Faraday::ParsingError` need to rescued somewhere else? 

**2: Not being able to use `slack-ruby-danger`**

The danger gem currently doesn't support Faraday 2.0 so for the moment I've commented it out. It has been [mentioned](https://github.com/slack-ruby/slack-ruby-client/issues/399#issuecomment-1110879664) about pulling it into its own `Gemfile`.


schinery:
Worth noting, I tried running the specs with the `CONCURRENCY` and `SLACK_API_TOKEN` set but was having issues with the Slack token. The rest of the specs ran successfully, so fingers crossed these will run ok with the correct setup.

dblock:
> I believe this is caused by there being no middleware equivalent to `::FaradayMiddleware::ParseJson`, and the new `:json` request/response middleware appears to only parse (and throw `::Faraday::ParsingError`) if the content type is `application/json`.

That makes a lot of sense. Change the spec to supply `application/json`.

> So, I've added some code to check if the `env.body` has been converted into `Hash` or not, and attempt to convert if it isn't.

It's probably good to be defensive and not reach into the hash unless it's a hash, but don't attempt to convert it. Add a separate spec with the correct content-type and with a different one.
 
> It does raise the question does `::Faraday::ParsingError` need to rescued somewhere else?

Probably yes, and raise `Slack::Web::Api::Errors::ParsingError` that would wrap that one?
 
> **2: Not being able to use `slack-ruby-danger`**
> 
> The danger gem currently doesn't support Faraday 2.0 so for the moment I've commented it out. It has been [mentioned](https://github.com/slack-ruby/slack-ruby-client/issues/399#issuecomment-1110879664) about pulling it into its own `Gemfile`.

Yep, I did it in https://github.com/slack-ruby/slack-ruby-bot-server/pull/142/files#diff-9ed6d3c8f52cf7b6862dd9628741eda3dbcd3683b1485977c483b6571e989904 if you just want to copy-paste.

Gemfile
```diff
@@ -15,13 +15,15 @@ end
 group :test do
   gem 'activesupport'
   gem 'erubis'
+  gem 'faraday-typhoeus'
```
dblock, line 18: Is this becoming a gem dependency? Does it belong in .gemspec?

lib/slack/web/faraday/response/raise_error.rb
```diff
@@ -3,28 +3,33 @@ module Slack
...
-          def call(env)
-            super
-          rescue ::Faraday::ParsingError
-            raise Slack::Web::Api::Errors::ParsingError, env.response
+          private
+
+          # Need to check if the body has been parsed here as it will only be parsed by the Faraday JSON middleware if
+          # the response is content_type 'application/json'. This would of been handled automatically by the
+          # ::FaradayMiddleware::ParseJson but that is no longer available.
+          def process_body(env)
```
dblock, line 31: Sounds like this is no longer needed because Faraday does it.

dblock (changes requested):
Good progress. More comments below.

Also update CHANGELOG, README, and remove incompatible versions of Ruby from CI.

schinery:
> Also update CHANGELOG, README, and remove incompatible versions of Ruby from CI.

Ah this answers my Ruby version question above 👍🏻

Gemfile
schinery: This was added as testing dependency because it was required for the 'non default adapter' spec in `spec/slack/web/client_spec.rb:177` and the adapters are no longer bundled in.

CHANGELOG.md
```diff
@@ -3,7 +3,9 @@
 * [#400](https://github.com/slack-ruby/slack-ruby-client/pull/400): Replace Travis-CI with GitHub Actions - [@dblock](https://github.com/dblock).
 * [#401](https://github.com/slack-ruby/slack-ruby-client/pull/401): Upgraded RuboCop to 1.26.0 - [@dblock](https://github.com/dblock).
 * [#401](https://github.com/slack-ruby/slack-ruby-client/pull/401): Added support for Ruby 3.1.1 - [@dblock](https://github.com/dblock).
-* Your contribution here.
+* [#406](https://github.com/slack-ruby/slack-ruby-client/pull/406): Update Faraday to `>= 2.0` - [@schinery](https://github.com/slack-ruby/slack-ruby-client/pull/406).
```
dblock, line 8: Let's bump the version to 1.1 (version.rb, etc.)

lib/slack/web/faraday/response/raise_error.rb
```diff
@@ -3,28 +3,33 @@ module Slack
...
 
-            error_message =
-              body['error'] || body['errors'].map { |message| message['error'] }.join(',')
-
+            error_message = body['error'] || body['errors'].map { |message| message['error'] }.join(',')
             error_class = Slack::Web::Api::Errors::ERROR_CLASSES[error_message]
             error_class ||= Slack::Web::Api::Errors::SlackError
             raise error_class.new(error_message, env.response)
+          # TODO: Do we need to still handle ::Faraday::ParsingError or will this be handled else where?
+          # rescue ::Faraday::ParsingError
```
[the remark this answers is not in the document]
schinery: This parse rescue has been moved to `lib/slack/web/faraday/request.rb` as the `Faraday::ParsingError` is now triggered when processing the response.

As part of this, and to handle non JSON responses such as HTML 500s, I've currently set `content_type: /\b*$/` in the `:json` response options. The question is do you think this should be more restrictive and only parse certain known content types?

lib/slack/web/faraday/response/raise_error.rb
schinery: Yeah this wasn't needed after all and has been removed.

spec/slack/web/client_spec.rb
```diff
@@ -280,8 +281,7 @@
         it 'raises ParsingError' do
```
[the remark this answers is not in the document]
schinery: Updated the spec that checked for the HTML response and added one for malformed JSON. 

There is the [question above](https://github.com/slack-ruby/slack-ruby-client/pull/406#discussion_r864463221) regarding correct content types as the new code will currently try and parse any content type returned, which I guess these two specs cover?

schinery:
@dblock I think this is ready to be re-reviewed as I've...

* Resolved the JSON parsing stuff previously discussed. There is [this question](https://github.com/slack-ruby/slack-ruby-client/pull/406#discussion_r864463221) which based on the answer might need some additional changes
* Created `Gemfile.danger` and set the GitHub action to use that so Faraday 1.0 doesn't affect the main `bundle install`
* Updated the `CHANGELOG`, although let me know if I've gone a bit overboard here

One thing, I couldn't see what might need updating in the `README` except maybe updating the `Stable Release` section. Let me know if this is what you meant.

.github/workflows/test.yml
```diff
@@ -6,12 +6,10 @@ jobs:
     strategy:
       matrix:
         entry:
-          - { ruby: 2.5.3 }
-          - { ruby: 2.6.2 }
-          - { ruby: 2.7.1 }
-          - { ruby: 2.7.1, concurrency: async-websocket }
-          - { ruby: 3.0.2 }
-          - { ruby: 3.1.1 }
+          - { ruby: 2.7 }
```
dblock, line 9: Minor: quote all rubies so we don't make the 3.0 being a float mistake

CHANGELOG.md
```diff
@@ -1,11 +1,19 @@
+### 1.1.0 (TBC)
+
+* [#406](https://github.com/slack-ruby/slack-ruby-client/pull/406): Update Faraday to `>= 2.0` - [@schinery](https://github.com/slack-ruby/slack-ruby-client/pull/406).
+* [#406](https://github.com/slack-ruby/slack-ruby-client/pull/406): Removed support for Ruby 2.5 and 2.6 - [@schinery](https://github.com/slack-ruby/slack-ruby-client/pull/406).
+* [#406](https://github.com/slack-ruby/slack-ruby-client/pull/406): Test against latest Ruby 2.7, 3.0 and 3.1 - [@schinery](https://github.com/slack-ruby/slack-ruby-client/pull/406).
+* [#406](https://github.com/slack-ruby/slack-ruby-client/pull/406): Upgraded RuboCop to 1.26.1 and lock to version - [@schinery](https://github.com/slack-ruby/slack-ruby-client/pull/406).
+* [#406](https://github.com/slack-ruby/slack-ruby-client/pull/406): Updated Danger PR Linter GitHub action to use separate `Gemfile`  - [@schinery](https://github.com/slack-ruby/slack-ruby-client/pull/406).
+* Your contribution here.
+  
 ### 1.0.1 (Next)
```
dblock, line 10: This hasn't shipped, merge with the above.

CHANGELOG.md
```diff
@@ -1,11 +1,19 @@
...
 
 * [#400](https://github.com/slack-ruby/slack-ruby-client/pull/400): Replace Travis-CI with GitHub Actions - [@dblock](https://github.com/dblock).
 * [#401](https://github.com/slack-ruby/slack-ruby-client/pull/401): Upgraded RuboCop to 1.26.0 - [@dblock](https://github.com/dblock).
 * [#401](https://github.com/slack-ruby/slack-ruby-client/pull/401): Added support for Ruby 3.1.1 - [@dblock](https://github.com/dblock).
 * [#405](https://github.com/slack-ruby/slack-ruby-client/pull/405): Added `admin_apps_requests_cancel`, `admin_users_unsupportedVersions_export`, `bookmarks_add`, `bookmarks_edit`, `bookmarks_list`, `bookmarks_remove` - [@dblock](https://github.com/dblock).
 * [#404](https://github.com/slack-ruby/slack-ruby-client/pull/404): Update `chat_update` method not to raise error with parameter which contains only `reply_broadcast` - [@colorbox](https://github.com/colorbox).
-* Your contribution here.
```
dblock, line 8: Put this one back.

lib/slack/web/faraday/connection.rb
```diff
@@ -22,11 +22,14 @@ def connection
...
+                connection.request :multipart
+                connection.request :url_encoded
                 connection.use ::Slack::Web::Faraday::Response::RaiseError
-                connection.use ::FaradayMiddleware::Mashify, mash_class: Slack::Messages::Message
-                connection.use ::FaradayMiddleware::ParseJson
+                connection.response :mashify, mash_class: Slack::Messages::Message
+                # Match any content type for JSON responses meaning Faraday::ParsingError will be raised for non-JSON
+                # responses such as HTML 500 errors.
+                # TODO: Should this be more restrictive?
+                connection.response :json, content_type: /\b*$/
```
dblock, line 32: So this is any content type, isn't it? Do any specs fail if you remove `content_type` here? I think not, and it's the old behavior that basically says "try to parse anything, which now raises ::Faraday::ParsingError, which we handle".

dblock (changes requested):
I think we're almost there! Made a bunch of minor comments and try to address the content-type one below.

lib/slack/web/faraday/response/raise_error.rb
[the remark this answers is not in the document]
dblock: This means content-type can be anything, doesn't it? So what's the point in having it? I think we should just try to parse any response because the Slack API says it will always send JSON, except when it's broken, and that's when we need this.

lib/slack/web/faraday/connection.rb
schinery: If we remove this `content_type:` option then the `<html></html>` response specs will fail because the default for the [JSON middleware](https://lostisland.github.io/faraday/middleware/json-response) parsing is `/\bjson$/`, so we wouldn't attempt to parse if a 500 HTML response was returned and we end up with the `undefined method 'map' for nil:NilClass` mention in the initial PR description.

I guess there are two options:

1) Leave as is it now meaning that it will try and parse any response Slack returns, which I think as you say is the old behaviour handled by the deprecated `::FaradayMiddleware::ParseJson` middleware
2) Make the content types stricter and set it to only attempt to parse JSON or HTML content types as we can set an array of content types

schinery: > Remove comment and TODO above.

Removed 👍🏻

dblock (approved):
This looks good! Merging. Let's see what falls out in integration tests run on master with a token. My local run with `CONCURRENCY=async-websocket SLACK_API_TOKEN=... bundle exec rake` runs integration tests successfully.

dblock:
Btw, if you are bored, I have a few more clients that could use some love of upgrading faraday, https://github.com/dblock/strava-ruby-client, https://github.com/dblock/iex-ruby-client, and https://github.com/dblock/open-weather-ruby-client.

PR #410, opened 2022-06-07
Suppress deprecation warnings when running tests on the CI

schinery:


dblock:
@schinery I wonder whether the deprecation warnings are going to stderr and the expected output to stdout?

Issue #415, opened 2022-07-19
SSL Error on Heroku-22: SSL_CTX_load_verify_file: system lib


irphilli:
On [Heroku-22](https://devcenter.heroku.com/articles/heroku-22-stack) (based on Ubuntu 22), I get the following error when trying to use the client:

```
SSL_CTX_load_verify_file: system lib
```

Something as simple as this causes the error:
```
Slack::Web::Client.new(token: 'abc').auth_test
```

Doing the following is a workaround:
```
Slack::Web::Client.new(token: 'abc', ca_file: nil, ca_path: nil).auth_test
```

Looking at the source, it looks like the gem is defaulting to `OpenSSL::X509::DEFAULT_CERT_FILE` and `OpenSSL::X509::DEFAULT_CERT_DIR`. On Heroku-22, at least, these paths aren't valid.

Any reason for defaulting to those, as opposed to `nil`? (i.e. let OpenSSL figure out the paths).

dblock:
No good reason AFAIK. I say we either document it or change it with a major version bump. Care to PR either?

PR #416, opened 2022-07-20
Removes default values for Faraday's SSL settings `ca_file` and `ca_path`

irphilli:
Fixes: #415 

Can alternatively just blanket default to `nil`, but this seems like this would be a bit less disruptive in case anyone was relying on the current defaults.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
-### 1.1.1 (Next)
+### 2.0.0 (Next)
 
+* [#416](https://github.com/slack-ruby/slack-ruby-client/pull/416): Removes default values for Faraday's SSL settings `ca_file` and `ca_path` - [@irphilli](https://github.com/irphilli).
 * Your contribution here.
 
 ### 1.1.0 (2022/06/05)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ### 1.1.1 (Next)
 
-* Your contribution here.
```
dblock, line 3: Put this back.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
-### 1.1.1 (Next)
+### 2.0.0 (Next)
 
+* [#416](https://github.com/slack-ruby/slack-ruby-client/pull/416): Removes default values for Faraday's SSL settings `ca_file` and `ca_path`. - [@irphilli](https://github.com/irphilli).
```
dblock, line 3: You have an extra `.` mid sentence, PR bot is unhappy.

dblock:
@irphilli Take a look at https://github.com/slack-ruby/slack-ruby-client/issues/418 just to confirm that we have't created a new issue, pls?

PR #417, opened 2022-07-26
Raise rescuable errors

zachahn:
This commit replaces all instances of `throw` with `raise`.

While I was using this library, I noticed that error handling didn't
work as expected.

Here's an example of what I tried. Looking at the code, I expected this
to print `ArgumentError`, but it printed `UncaughtThrowError`.

```ruby
client = Slack::Web::Client.new(token: ENV["TOKEN"])
begin
  client.chat_postMessage
rescue => e
  puts e.class
end
# UncaughtThrowError
```

According to Ruby's [tutorial on Exceptions][1], `raise` and `rescue`
should be used for exceptions, while `throw` and `catch` could be used
for control flow.

[1]: https://ruby-doc.com/docs/ProgrammingRuby/html/tut_exceptions.html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### 2.0.0 (Next)
 
 * [#416](https://github.com/slack-ruby/slack-ruby-client/pull/416): Removes default values for Faraday's SSL settings `ca_file` and `ca_path` - [@irphilli](https://github.com/irphilli).
+* [#417](https://github.com/slack-ruby/slack-ruby-client/pull/417): Raise rescuable errors - [@zachahn](https://github.com/zachahn).
 * Your contribution here.
 
 ### 1.1.0 (2022/06/05)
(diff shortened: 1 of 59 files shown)


dblock:
Fair enough, thanks! 

Please add a line to CHANGELOG, get CI to pass, there's a bunch of Rubocop suggestions.

zachahn:
Done and done. Since I modified some patch files, I verified my changes by running `bundle exec rake slack:api:update` (I didn't commit those changes).

Let me know if there's anything else is missing. Thanks!

dblock:
> Done and done. Since I modified some patch files, I verified my changes by running `bundle exec rake slack:api:update` (I didn't commit those changes).

Looks like they are committed tho (those files that have "This file was auto-generated by lib/tasks/web.rake" on top). I would either include all those generated changes, or include none of the generated changes (and make a separate PR/commit with those)?

dblock:
@zachahn Want to followup with other API updates and we can make a release?

PR #419, opened 2022-10-02
Remove rtm.start and start_method configuration

kstole:
Now that the `rtm.start` method has been deprecated (and has a response matching `rtm.connect`), there is only one method for connecting to the Real Time Messaging API.


kstole:
We get a lot of changes by running the API update tasks, so I left those changes out of this and just removed the `#rtm_start` method manually

.rubocop.yml
```diff
@@ -35,6 +35,10 @@ Style/Documentation:
 Style/ModuleFunction:
   Enabled: false
 
+Style/NumericLiterals:
+  Exclude:
+    - 'spec/**/*'
```
kstole, line 40: I'm not sure the best way to disable this just for Unix timestamps so I just disabled it entirely for specs

dblock:
Good stuff, thanks!

dblock:
Added you here too to maintainers (https://github.com/slack-ruby/slack-ruby-client/invitations), thanks for picking up the overdue update tasks @kstole!

PR #420, opened 2022-10-04
Fix default store missing event hooks

kstole:
The default Slack store (`Slack::RealTime::Store`) inherits from `Slack::RealTime::Stores::Store` but didn't have access to its event hooks stored in a class instance variable of the superclass. This ensures store subclasses inherit the event hooks from their inheritance chain.

### DIFF
--- a/lib/slack/real_time/client.rb
+++ b/lib/slack/real_time/client.rb
@@ -11,12 +11,6 @@ module Slack
       include Api::Message
       include Api::Typing
 
-      @events = {}
-
-      class << self
-        attr_accessor :events
-      end
-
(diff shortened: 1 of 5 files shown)


lib/slack/real_time/client.rb
```diff
@@ -11,12 +11,6 @@ class ClientAlreadyStartedError < StandardError; end
       include Api::Message
       include Api::Typing
 
-      @events = {}
-
-      class << self
-        attr_accessor :events
-      end
```
kstole, line 18: This was left unused when event hooks were migrated to inside the stores (https://github.com/slack-ruby/slack-ruby-client/commit/d612c35ee7b8f59a68a35f40a20fc3c296c77cd1)

PR #423, opened 2022-10-10
Initialize store data on successful connection (hello)

kstole:
This is based on the fix my team implemented in our bot to replace the old `rtm.start` behavior. I debated where this belongs and open to feedback. One option would be to put it into `Slack::RealTime::Client#build_socket` and `Slack::RealTime::Client#restart_async`, which could retrieve data from the Web API before _or_ after connecting to the RTM API, but it seemed cleaner triggering it on the successful `hello` message. It could also be a standalone built-in handler (that's how we implemented in our bot), but it makes a lot of sense belonging to the store and automatically avoids conflicting with the `Starter` store. Finally, I considered an option to disable this initialization functionality, but that's essentially possible by having it in `Slack::RealTime::Store` and not `Slack::RealTime::Stores::Store` (detailed in README and UPGRADING).


dblock:
I am a fan of doing less by default and letting users do expensive things like enumerating all users knowingly. I can't imagine how long `users_list` or `conversations_list` is going to take on a large corporate slack. The reason why `rtm.start` was deprecated is precisely for the same reason why we should deprecate the initialization behavior here.

So I am thinking we should do this:

1. By default I would keep `Slack::RealTime::Store` as is.
2. Document in UPGRADING.md that since https://github.com/slack-ruby/slack-ruby-client/pull/419/files, and because of the `rtm.start` change, the store is no longer initialized with data.
3. Introduce a new store, `Slack::RealTime::Auto` that initializes channels, users, etc.
4. [Optional] Make `Slack::RealTime::Auto` configurable for what it is capable of enumerating (channels vs. users).
5. Document in UPGRADING.md to use `Slack::RealTime::Auto` to obtain previous `rtm.start` behavior, and document differences.

WDYT?

kstole:
My thought was trying to keep the previous default behavior of the gem (using `rtm.start` and `Slack::RealTime::Store` fully loaded) and I think this is closest to the previous behavior (at least it paginates the data loading). There's a difference, though, that setting `start_method` was a way to change the store class (if you chose `rtm_connect`, you got the `Starter` store by default). Now that `rtm_connect` is the only start method, it seems strange to use the full `Store` that reacts to all changes if it doesn't have that data loaded.

I'm fine with creating a new store that houses this additional initialization but then I think `Starter` should be the default store. And at that point, what are we doing with `Slack::RealTime::Store`?

dblock:
> My thought was trying to keep the previous default behavior of the gem (using `rtm.start` and `Slack::RealTime::Store` fully loaded) and I think this is closest to the previous behavior (at least it paginates the data loading). 

Yes, I understand. But this is the time to tell people to stop doing that since that kind of behavior was deprecated. You're offering a path back, but it shouldn't be the default as a major breaking change because it's bad (TM).

> There's a difference, though, that setting `start_method` was a way to change the store class (if you chose `rtm_connect`, you got the `Starter` store by default). Now that `rtm_connect` is the only start method, it seems strange to use the full `Store` that reacts to all changes if it doesn't have that data loaded. I'm fine with creating a new store that houses this additional initialization but then I think `Starter` should be the default store. And at that point, what are we doing with `Slack::RealTime::Store`?

Yes! Right on. We can probably just remove `Slack::RealTime::Store`?

lib/slack/real_time/stores/loaded.rb
```diff
@@ -0,0 +1,80 @@
...
+            end
+          end
+
+          if caches.include?(:channels) || caches.include?(:groups) || caches.include?(:ims) || caches.include?(:mpims)
+            types = CONVERSATION_TYPES.slice(*caches).values.join(',')
+            client.web_client.conversations_list(types: types) do |conversations|
+              conversations.channels.each do |channel_data|
+                if channel_data.is_mpim?
+                  mpim = Models::Mpim.new(channel_data)
+                  mpims[mpim.id] = mpims.key?(mpim.id) ? mpims[mpim.id].merge(mpim) : mpim
```
kstole, line 62: I don't think there's much use for mpims in a bot, but I added it as an option for completeness and because it seems like mpims (that include the bot user) had previously been included in the list of channels returned in the `rtm.start` response (they're considered a type of channel)

lib/slack/real_time/stores/loaded.rb
```diff
@@ -0,0 +1,80 @@
...
+                elsif channel_data.is_im?
+                  im = Models::Im.new(channel_data)
+                  ims[im.id] = ims.key?(im.id) ? ims[im.id].merge(im) : im
+                elsif channel_data.is_group?
+                  group = Models::Group.new(channel_data)
+                  groups[group.id] = groups.key?(group.id) ? groups[group.id].merge(group) : group
+                elsif channel_data.is_channel?
+                  channel = Models::Channel.new(channel_data)
+                  channels[channel.id] = channels.key?(channel.id) ? channels[channel.id].merge(channel) : channel
```
kstole, line 71: I'm conflicted about what to do with groups. I distinctly remember when I set up a bot several years ago that `channels` were effectively public channels and `groups` were private channels.

I finally found this explanation: https://api.slack.com/types/group
> A group object contains information about a private channel created before March 2021. The private channel's ID will begin with G, as private channels were once known as "private groups." [Conversation objects](https://api.slack.com/types/types/conversation) contain the information for private channels created after March 2021 (those with IDs that begin with C).

I'm not sure why they changed it, but my interpretation is that groups aren't created anymore.

So there are
public channels: `{ is_channel: true, is_group: false, is_private: false }`
private channels (created after March 2021): `{ is_channel: true, is_group: false, is_private: true }`
private channels (created before March 2021): `{ is_channel: false, is_group: true, is_private: true }`

It seems strange to me that groups are not labeled as `is_channel`, but they are effectively treated the same and `conversations_list(types: 'private_channel')` returns both private channel types in the same list.

I think having separate caches for the public channels and private channels is a lot more useful than a cache for all "channels" (both public and private) and a separate cache for just groups, but I'm not sure what to name them. Keeping them named `channels` and `groups` is convenient and probably makes sense to someone familiar with the pre-March 2021 behavior, but calling them `channels`/`public_channels` and `private_channels`, while more verbose, may be clearer for fresh users.

kstole (commented):
@dblock I still need to clean up some stuff and finish specs, but wanted to get your feedback on this approach

lib/slack/real_time/stores/loaded.rb
dblock: Sigh, this stuff is wild :) I think we should remove any legacy behavior altogether, thus calling `conversations_list` with `public_channel` on startup by default, making it configurable, and storing a hash by `type` from the new API, so only buckets like `public_channel`, `private_channel`, `mpim`, `im` that match 100% what you'd get back from the API if you specified `type`. WDYT?

kstole: I agree with matching the store's caches up with the types used by Slack (despite `public_channels` and `private_channels` being annoyingly long) 😬 

When you say
> storing a hash by type from the new API ... extensible for any future types Slack might add

are you thinking the stores only contain a single variable with a Hash that gets the types as keys? I prefer still having a separate Hash for each type like it has now. I think that's simpler and easier for dealing with the types as needed and the types are so fundamental that they rarely change (private groups becoming private channels is the only change in the last few years AFAIK).

And then I was also unsure about this
> thus calling `conversations_list` with `public_channel` on startup by default

since before you mentioned doing less by default. I guess if all of the extra initialization calls are configurable, there isn't a need for the separate `Loaded` store I set up and they could instead just be in the main `Store`. It's just a question of which caches should be fetched by default. Could be none, could be all, or could be ones we consider typically useful (like all except `mpims`). If `Starter` is the default store, then there's some flexibility in the default configuration for the standard `Store`.

dblock: From the developer perspective I think I prefer to write `store.conversations.public_channel` vs. `store.public_channel`, because it matches the API, but that's not the hill I want to die on.

Agree on fetching nothing by default, I think you're right. As long as it's configurable.

kstole: Ah I see what you were saying now with `store.conversations.public_channel`. My counter would be that the methods have changed more frequently than the types so I think it's important to decouple them. There have also been different methods that could fetch similar data and already we would be using a few different methods for getting team data vs user-like data vs channel-like data.

If that's an acceptable approach, I'll try to update this by today or tomorrow :)

dblock: Acceptable!

lib/slack/real_time/stores/store.rb
```diff
@@ -281,29 +302,80 @@ def initialize(attrs)
...
         # The client has successfully connected to the server.
         # @see https://api.slack.com/events/hello
         # @see https://github.com/slack-ruby/slack-api-ref/blob/master/events/hello.json
-        # on :hello do |data|
+        on :hello do |_data, client|
+          # rubocop:disable Metrics/BlockNesting
+          if caches.include?(:teams)
+            team_info = client.web_client.team_info
+            team = Models::Team.new(team_info.team)
+            teams[team.id] = teams.key?(team.id) ? teams[team.id].merge(team) : team
```
kstole, line 333: The old `rtm.start` response included `prefs` for the team. There’s a `client.web_client.team_preferences_list`, but despite what https://api.slack.com/methods/team.preferences.list says, it doesn’t seem to work with a legacy bot token. I get a `missing_scope` error. Which is strange because that token does allow the store to receive updates to individual prefs through the `team_pref_change` event.

README.md
```diff
@@ -38,11 +38,12 @@ A Ruby client for the Slack [Web](https://api.slack.com/web), [RealTime Messagin
...
-    - [Slack::RealTime::Stores::Starter](#slackrealtimestoresstarter)
-  - [Configuring Slack::RealTime::Client](#configuring-slackrealtimeclient)
-    - [Caveats](#caveats)
-      - [websocket_ping](#websocket_ping)
+    - [Configuring Slack::RealTime::Client](#configuring-slackrealtimeclient)
+      - [Caveats](#caveats)
+        - [websocket_ping](#websocket_ping)
+    - [RealTime Store](#realtime-store)
+      - [Slack::RealTime::Stores::Starter](#slackrealtimestoresstarter)
+      - [Slack::RealTime::Stores::Store](#slackrealtimestoresstore)
```
kstole, line 46: I could see renaming these to something like `Light` and `Full` to be more clear, since `Store` along isn't very clear, but I don't know if there's much gain.

UPGRADING.md
```diff
@@ -20,7 +20,14 @@ or
...
+If your app still relies on the old `rtm.start` behavior, you can achieve a similar result with `Slack::RealTime::Stores::Store`, which initializes itself using Web API methods (see [README](README.md#slackrealtimestoresstore)).
+
+The names of the Store caches `channels` and `groups` have been changed to `public_channels` and `private_channels` to reflect their true meaning.
+This is consistent with the [Slack Web API](https://api.slack.com/methods/conversations.list#arg_types), although [event types for private channels](https://api.slack.com/events?query=group) still use the term `group`.
+In both cases, private channels consist of both [conversation types](https://api.slack.com/types/conversation) (`id: 'C...'', is_channel: true, is_group: false, is_private: true`) and [group types](https://api.slack.com/types/group) (`id: 'G...', is_channel: false, is_group: true, is_private: true`) depending whether the object was created before or after March 2021.
```
kstole, line 30: Slack throws in a bit of a wrinkle where it still uses `group_` for events. Given that there's consistency in separating private channels from public channels, regardless of whether they're considered a channel or group, I think the new split in the Store makes a lot of sense. Also, not all of the Slack docs have been updated to reflect that private channels can be either type.

lib/slack/real_time/stores/store.rb
dblock: 🤷 generally I just comment things like these out and open another issue

lib/slack/real_time/stores/store.rb
```diff
@@ -108,20 +111,29 @@ def initialize(attrs)
...
+              private_channels[channel_id] = Models::Channel.new(data.channel)
+            end
           else
-            channels[channel_id] = Models::Channel.new(data.channel)
+            channel = public_channels[channel_id]
+            if channel
+              channel.merge!(data.channel)
+            else
+              public_channels[channel_id] = Models::Channel.new(data.channel)
+            end
```
dblock, line 127: This logic would be a lot simpler and future proof if we just stored `channels[]`. The implementation for `private_channels` would be `channels.select(&:is_private)`. The additional notifications such as renames and unarchives are for all channels AFAIK as well. I'll merge as is, but consider this change before we release?

lib/slack/real_time/stores/store.rb
```diff
@@ -108,20 +111,29 @@ def initialize(attrs)
...
           end
         end
 
         # You left a channel.
         # @see https://api.slack.com/events/channel_left
         # @see https://github.com/slack-ruby/slack-api-ref/blob/master/events/channel_left.json
         on :channel_left do |data|
-          channel = channels[data.channel]
-          channel.members.delete(self.self.id) if channel.key? :members
+          channel = public_channels[data.channel]
```
dblock, line 135: Is this always a public channel? I'm not so sure. Same for other channel events.

lib/slack/real_time/stores/store.rb
```diff
@@ -236,21 +253,24 @@ def initialize(attrs)
...
+          channel = private_channels[data.channel]
           channel.is_archived = true if channel
         end
 
         # You closed a private channel.
         # @see https://api.slack.com/events/group_close
         # @see https://github.com/slack-ruby/slack-api-ref/blob/master/events/group_close.json
         on :group_close do |data|
-          groups[data.channel].is_open = false
+          channel = private_channels[data.channel]
```
dblock, line 264: Note that group events do say they are for private channels, but channel events do not say they are only for public ones (https://api.slack.com/events/channel_deleted).

lib/slack/real_time/stores/store.rb
```diff
@@ -281,29 +302,80 @@ def initialize(attrs)
...
+        on :hello do |_data, client|
+          # rubocop:disable Metrics/BlockNesting
+          if caches.include?(:teams)
+            team_info = client.web_client.team_info
+            team = Models::Team.new(team_info.team)
+            teams[team.id] = teams.key?(team.id) ? teams[team.id].merge(team) : team
+          end
+
+          if caches.include?(:users) || caches.include?(:bots)
+            client.web_client.users_list do |users_list|
```
dblock, line 337: Maybe extract into a method?

dblock (approved):
I'm merging this. Let's keep iterating for any leftovers.

lib/slack/real_time/stores/store.rb
kstole: Yeah I noticed that. The Slack docs are incomplete in areas. I tested most of these events and observed only public channels triggering the `channel_` events while only private channels (both newer and group type) triggering the `group_` events.

lib/slack/real_time/stores/store.rb
kstole: https://github.com/slack-ruby/slack-ruby-client/pull/423#discussion_r998352545

Issue #426, opened 2022-10-18
Integration tests broken


dblock:
Looks like we broke integration tests that only run with a token on master.

https://github.com/slack-ruby/slack-ruby-client/actions/runs/3274143367/jobs/5387272594

cc: @kstole

kstole:
I'm not sure what exactly broke the test (if I'm reading correctly, seems like something with `console/progress.rb` in `webmock/rspec`) but I was able to reproduce locally by switching to Ruby 2.7.1 (it passes on 2.7.6). I think we can just update to the latest 2.7 patch.

dblock:
@kstole Right on. It's expecting some console output and that now spit other things out. Closing via https://github.com/slack-ruby/slack-ruby-client/pull/427.

PR #427, opened 2022-10-18
Update GitHub Actions to latest versions

kstole:
Using the latest version of [actions/checkout](https://github.com/marketplace/actions/checkout) will resolve the `Node.js 12 actions are deprecated` warnings.

Ruby version in `.ruby-version` is used by [ruby/setup-ruby](https://github.com/marketplace/actions/setup-ruby-jruby-and-truffleruby) automatically when not otherwise specified.

Addresses #426


.github/workflows/lint.yml
```diff
@@ -1,13 +1,14 @@
 name: Rubocop
```
kstole, line 1: I wanted to rename this workflow to `Lint` to match [slack-api-ref](https://github.com/slack-ruby/slack-api-ref/blob/master/.github/workflows/lint.yml) and so we could add other linters to this workflow if needed, but GitHub wouldn't recognize the history of previous runs. Maybe we can rename this in the future.

.github/workflows/integration_test.yml
```diff
@@ -1,19 +1,21 @@
...
+on: push
 jobs:
   test:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         entry:
-          - { ruby: 2.7.1, concurrency: async-websocket }
+          - { ruby: 2.7, concurrency: async-websocket }
     name: test (ruby=${{ matrix.entry.ruby }}, concurrency=${{ matrix.entry.concurrency || 'none' }})
```
kstole, line 10: It's strange to me that this is using the matrix strategy for a single run. Idk if it was just copied from the standard test workflow or if it was intentional. We could set it up like other simple workflows to pull ruby version from `.ruby-version`

dblock: It used to have multiple concurrency models, faye-websocket and async-websocket, thus it was a matrix. Then the matrix got simpler. You can definitely collapse this if you want to.

.github/workflows/lint.yml
dblock: No strong opinions.

dblock (approved):

PR #434, opened 2022-11-27
fix GLI usage breaking faraday >= 2.7.0

### LINKED ISSUE #433: Incompatibility between Faraday >= 2.7.0 and bin/slack script including GLI::App at the top level
**Steps to Reproduce**
- Have faraday >= 2.7.0 installed
- Run the `bin/slack` command in some way that would produce an error
  - `slack --vcr-cassette-name=web/auth_test_error --slack-api-token=token auth test` from the test suite will do

**Observed behavior**
You get an error message: `error: wrong number of arguments (given 1, expected 0)`

**Desired behavior**
You get the actual error message, e.g. `error: not_authed`

**Explanation**
- The `slack` script [includes `GLI::App` at the top level](https://github.com/slack-ruby/slack-ruby-client/blob/e6c1a3f2ff95bf54409d92757ff86583031119ad/bin/slack#L5)
  - This unfortunately includes `GLI::App` in the `Object` class:
    ```
    $ ruby -e 'module M; end; include M; puts Object.ancestors.inspect'
    [Object, M, Kernel, BasicObject]
    ```
- [`Faraday::Middleware` calls `on_error` on itself with an exception it that method is implemented](https://github.com/lostisland/faraday/blob/c6668ef8a8dc16113d6a4d1c7e88521413b7e4eb/lib/faraday/middleware.rb#L21)
- This resolves to [`GLI::App#on_error` which has a completely different meaning](https://github.com/davetron5000/gli/blob/8498d40346e0340654873d5b4cb728e31c262f8e/lib/gli/app.rb#L162-L164)

**Possible Fixes**
- replace `include GLI::App` with `extend GLI::App` ... all the tests seem to pass with this change
- Wrap the whole CLI in a class that has `extend GLI::App` (this seems to be what GLI's docs mostly reference)

blowfishpro:
when using include at the top level, the module gets included in the Object class

`Faraday::Middleware` checks whether it responds to `on_error`, which then finds GLI's `on_error`, which has a different signature and meaning

this resolves that by wrapping the entire CLI in a class so that `GLI::App` is not included in `Object`

we also use GLI's built-in `commands_from` method to avoid having to reference every command file individually

Resolves #433

Note: The `slack:web:api:update` task seemed to want to pull in some new stuff - since that usually seems to be done separately I have intentionally excluded that but can include it if you want


dblock (changes requested):
These files *are* auto-generated to keep up with the API, so you want to modify the .erb template and re-generate them, not delete the template. You can do a 1-time update separately so that we can have something clean for this PR, let's start there?

blowfishpro:
That's what I did, I just did it while the code to update the submodule was commented out.

What documentation updates are required when new API features are pulled in?  I don't really understand the scope of changes that come with updating the submodule.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ### 2.0.1 (Next)
 
 * Your contribution here.
+* [#434](https://github.com/slack-ruby/slack-ruby-client/pull/434): Fix incompatibility with Faraday >= 2.7.0 and GLI usage in the `slack` script [@blowfishpro](https://github.com/blowfishpro)
```
dblock, line 4: This needs to match the format of other entries, it's missing a `- before your github account, and a period at the end.

dblock (changes requested):
Looks good! Danger is about to complain re:CHANGELOG format, could you please fix and I'll merge?

dblock:
> What documentation updates are required when new API features are pulled in? I don't really understand the scope of changes that come with updating the submodule.

Only CHANGELOG. You can leave that alone for now.

blowfishpro:
Fixed!

dblock:
FYI I decided to rename `CLI` to `Cli` in https://github.com/slack-ruby/slack-ruby-client/pull/435. LMK if you think it's a bad idea.

blowfishpro:
I'm pretty indifferent on that particular point as long as it's consistent.

PR #437, opened 2022-12-20
chore: replace danger token due to rotation

crazyoptimist:


crazyoptimist:
I suspect `secrets.SLACK_API_TOKEN` is also expired, seeing the failing [test logs](https://github.com/crazyoptimist/slack-ruby-client/actions/runs/3742377900/jobs/6353243679)

crazyoptimist:
Never mind, I'm wrong. It tries to find the token from my fork.

dblock:
> Never mind, I'm wrong. It tries to find the token from my fork.

Integration tests run on master here and have access to a token. We don't run those on PRs since they talk to actual Slack.

Issue #438, opened 2023-02-10
CLI error: wrong number of arguments (given 1, expected 0)


nello:
Hi, is this a dead project or am I missing some very important setup step? Just about all of the command-line functionality appears to be broken when I try it.

For example:

```
bundle exec slack users list
```

Gives the result:

```
error: wrong number of arguments (given 1, expected 0)
```

This is the same for pretty much every command except `bundle exec slack auth test`, being the only one that gives meaningful JSON output:

```
{"ok":true,"url":"https://myorg.slack.com/","team":"MyOrg","user":"techbot","team_id":"XXXXXXXXX","user_id":"XXXXXXXXX","bot_id":"XXXXXXXXX","is_enterprise_install":false}

```

Am I actually going insane?

dblock:
This smells like a faraday incompatibility that was fixed in https://github.com/slack-ruby/slack-ruby-client/pull/434 but not released yet (2.1.0). Can you please confirm with the code on master? I can do a release.

nello:
Hi, I can confirm that the latest CLI version on `master` doesn't give the same "wrong number of arguments" error. It seems to give new errors, though.

```
bundle exec ./slack users list
error: missing_scope
```

```
bundle exec ./slack channels info
error: Unknown command 'channels'
```

It also looks like it's expecting a very old 2.7.6 ruby, so possibly issues with parameter passing?

dblock:
The missing scope error means that the token you have doesn’t allow to call the API (not scoped to the API you’re calling). Read more about it here: https://api.slack.com/legacy/oauth-scopes

Channels were deprecated a while ago. If you found docs that mention channels we need to update them. To list channels you need to use “conversations”, https://api.slack.com/docs/conversations-api

duffn:
To note, you'll also get this error (on master) if you try something that the bot doesn't have permissions to do, for example post to a channel that it's not in. 

```
$ slack chat postMessage --text="hello world" --channel="#im-not-in-this-channel"
error: wrong number of arguments (given 1, expected 0)
```

dblock:
Confirmed fixed in 2.1.0 please reopen if you see otherwise.

PR #442, opened 2023-03-15
Update CLI channels docs

duffn:
Ref: https://github.com/slack-ruby/slack-ruby-client/issues/438#issuecomment-1427173940 The channels command has been deprecated and is not available in the CLI.

- The ID command is not available on the conversations method
- Change channels to conversations


dblock:
Thanks! Fix TOC pls?

PR #445, opened 2023-03-15
Only run integration tests in integration_test job

### LINKED ISSUE #444: Integration tests run more than integration tests & don't skip in GitHub Actions
There are two issues here:

- The integration test GitHub Action job https://github.com/slack-ruby/slack-ruby-client/blob/9d525298c5f30e0b6b80d6db5a1517467e66f593/.github/workflows/integration_test.yml runs all of the tests. See 467 examples run here: https://github.com/slack-ruby/slack-ruby-client/actions/runs/4427350917/jobs/7764813688
- The skip check in the integration test spec https://github.com/slack-ruby/slack-ruby-client/blob/9d525298c5f30e0b6b80d6db5a1517467e66f593/spec/integration/integration_spec.rb#L5 works locally but not in GitHub Actions. 

I can look at these.

duffn:
This:

1) skips the integration test job entirely in GitHub Actions if there isn't a Slack token in secrets.

It's particularly annoying to use the absence of a secret to control job flow. This is based on the answer here https://stackoverflow.com/questions/70249519/how-to-check-if-a-secret-variable-is-empty-in-if-conditional-github-actions

I went with the option of skipping the job entirely so it doesn't need to even checkout or install dependencies if the secret isn't there, but can be convinced this isn't the right approach.

Hidden secret, allows integration tests:

Screen Shot 2023-03-15 at 11 18 56 AM

No secret, skips integration tests:
Screen Shot 2023-03-15 at 11 19 11 AM

2) only runs the `integration` directory tests on this job

I'm not 100% sure why both `SPEC` and `SPEC_OPTS` are required here, but it doesn't work with either/or but only both

Closes #444

### DIFF
--- a/.github/workflows/integration_test.yml
+++ b/.github/workflows/integration_test.yml
@@ -1,8 +1,25 @@
 name: Integration Tests
 on: push
 jobs:
+  check-secret:
+    runs-on: ubuntu-latest
+    outputs:
+      secret-exists: ${{ steps.check-secret.outputs.defined }}
+    steps:
+      - name: Check for SLACK_API_TOKEN availability
+        id: check-secret
+        shell: bash
+        run: |
+          if [ "${{ secrets.SLACK_API_TOKEN }}" != '' ]; then
+            echo "defined=true" >> $GITHUB_OUTPUT;
+          else
+            echo "defined=false" >> $GITHUB_OUTPUT;
+          fi
+
   test:
     runs-on: ubuntu-latest
+    needs: [check-secret]
+    if: needs.check-secret.outputs.secret-exists == 'true'
     strategy:
       matrix:
         entry:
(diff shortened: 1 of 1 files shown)


.github/workflows/integration_test.yml
```diff
@@ -1,8 +1,25 @@
...
+          if [ "${{ secrets.SLACK_API_TOKEN }}" != '' ]; then
+            echo "defined=true" >> $GITHUB_OUTPUT;
+          else
+            echo "defined=false" >> $GITHUB_OUTPUT;
+          fi
+
   test:
     runs-on: ubuntu-latest
+    needs: [check-secret]
+    if: needs.check-secret.outputs.secret-exists == 'true'
```
dblock, line 22: ~Does a simple `if: ${{ secrets. SLACK_API_TOKEN }} != ''` not work? Do we need a whole extra step?~

I read up on this in https://stackoverflow.com/questions/70249519/how-to-check-if-a-secret-variable-is-empty-in-if-conditional-github-actions, sounds like this is the way :(

.github/workflows/integration_test.yml
```diff
@@ -1,8 +1,25 @@
 name: Integration Tests
 on: push
 jobs:
+  check-secret:
```
dblock, line 4: I'd like to nitpick on naming. Let's call this job `check-slack-api-token` and set `SLACK_API_TOKEN_EXISTS`?

duffn: Updated naming.

PR #448, opened 2023-03-17
Add JSON encoding to all applicable Web API arguments

jmanian:
There was at least one `chat.*` method that was missing the auto-encoding of `blocks` and `attachments` into JSON (`chat.scheduleMessage`). Instead of continuing to apply these to each method individually as a patch, I thought it would be nice to solve this once and for all.

~New groups of methods or new options that require encoding will require changes to accommodate those, but no patching.~ This is now being pulled from data in [slack-api-ref](https://github.com/slack-ruby/slack-api-ref/) and will update automatically with the docs.

~I'm open to suggestions on the exact approach, and I'm opening this as a draft without tests, changelog, etc. to gather comments first.~


lib/slack/web/api/templates/method.erb
```diff
@@ -51,6 +51,9 @@ module Slack
 <% if data['deprecated'] %>
             logger.warn('<%= group %>.<%= name %>: <%= data['deprecation']['deprecation_warning']%> Alternative methods: <%= data['deprecation']['alternative_methods'].join(', ')%>.')
 <% end %>
+<% if %w[chat views dialog].include?(group) %>
```
jmanian, line 54: I don't love that this array of strings will need to be updated in coordination with `encode_options` when new groups are added, but I also wasn't keen to add this line to all endpoints.

lib/slack/web/api/options.rb
```diff
@@ -0,0 +1,32 @@
...
+          when 'dialog'
+            encode_json(options, :dialog)
+          else
+            options
+          end
+        end
+
+        private
+
+        def encode_json(options, key)
```
dblock, line 21: Make `encode_json` take an array, so

```ruby
def encode_json(options, keys)
   Array(keys).each ...
     if ...
        options = 
     end
   end
   options
end
```

lib/slack/web/api/templates/method.erb
dblock: Can we promote `encoding: json` or a similar flag into slack-api-ref to avoid this?

dblock (commented):
I like this very much! See my suggestions/comments below.

jmanian:
@dblock Here's a different take, trying to base it entirely on data in `slack-api-ref`. It feels a little fragile to base it on data in the argument descriptions, but there isn't anything more structured at the moment (in `slack-api-ref` or on the doc pages). It does find all the ones we knew about, plus a handful of others.

jmanian:
@dblock still needs tests and changelog, but here's what this looks like now with the slack-ruby/slack-api-ref#64

dblock:
Looks good!

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ### 2.1.1 (Next)
```
dblock, line 1: Let's bump the version to 2.2 and make a release?

jmanian: Makes sense. I'm going to put up another PR to replace some of the patched argument validations with the new `arg_groups` data from the ref, so I'll do it in that.

PR #450, opened 2023-03-18
Add Jekyll docs

### LINKED ISSUE #443: Create documentation site
I'd like to create a documentation site for this repository. If this is not desired, please let me know and I can stop working on this. I've already begun.

_Thoughts_
- Use Jekyll and Just the Docs https://just-the-docs.github.io/just-the-docs/
- Translate all of the README to the docs site
- Add GitHub workflow to automatically publish the docs to GitHub pages at https://slack-ruby.github.io/slack-ruby-client (need repo owner help here to setup pages)
- I'll only directly copy the current documentation and then would plan to audit it for correctness after the migration

_Questions_
- There's a note in the current README how it's for HEAD and not for a particular version. This isn't currently available out of the box with Just the Docs https://github.com/just-the-docs/just-the-docs/issues/728. What to do here?
- I'm open to other themes or even other frameworks for docs, but Jekyll of course seemed the most appropriate and Just the Docs is a great theme

_Currently_

Screen Shot 2023-03-15 at 9 49 50 AM

Screen Shot 2023-03-15 at 9 50 09 AM

Screen Shot 2023-03-15 at 9 50 23 AM

duffn:
This adds docs built with Jekyll and Just the Docs per #443. You can view this on my fork here: https://duffn.github.io/slack-ruby-client/.

This will build the root docs on each push to master and the versioned tagged docs on each new tag creation. Head docs will be at https://slack-ruby.github.io/slack-ruby-client/ and versions docs at https://slack-ruby.github.io/slack-ruby-client/version/vX.XX/`

## Notes

- Will need to update the project's pages settings to look like this

Screen Shot 2023-03-17 at 7 10 04 PM

- Will need to update with each tag: 
  - [slack-ruby/slack-ruby-client@`master...duffn:slack-ruby-client:duffn/jekyll-docs?expand=1`#diff-b4d68dc855](https://github.com/slack-ruby/slack-ruby-client/compare/master...duffn:slack-ruby-client:duffn/jekyll-docs?expand=1#diff-b4d68dc855d0f9476d3f2ee343853bd21bf82ea9960d0cf06661baa244439dd6R25)
  - [slack-ruby/slack-ruby-client@`master...duffn:slack-ruby-client:duffn/jekyll-docs?expand=1`#diff-f0cbba616c](https://github.com/slack-ruby/slack-ruby-client/compare/master...duffn:slack-ruby-client:duffn/jekyll-docs?expand=1#diff-f0cbba616ce32d350d45cb6ae4c57fc71c505ac2a80e8c2b55264ccc54b3eb46R12)

- I haven't changed any doc content except when it fixed a broken link or removed some uncessary leading works in the case of adding the callout component like here: [slack-ruby/slack-ruby-client@`master...duffn:slack-ruby-client:duffn/jekyll-docs?expand=1`#diff-36f22022d6](https://github.com/slack-ruby/slack-ruby-client/compare/master...duffn:slack-ruby-client:duffn/jekyll-docs?expand=1#diff-36f22022d68d7e51443ece30e476272902b7e6d455bb8e1d9bd06bf1780f2b51R35-R36) (was "Note that the")

- If you want to push up v2.1.0 docs, you can build them locally and push to the `gh-pages` branch or take what I've already built here: https://github.com/duffn/slack-ruby-client/tree/gh-pages/version/v2.1.0

- I've tested and proofread as much as possible, but there's bound to be something missing or broken which I can fix forward.

Closes #443


dblock (commented):
I debated with myself whether this is a good idea. I enjoyed having a single README and not have to rely on an external website, but I am ok to part with it.

1. Is it possible to make version folders just `v[tag]`, I don't think we need the word `version` in the URL.
2. RELEASING.md needs to be updated. I believe one needs to add a line to the versions TOC? Or maybe we can generate that file by enumerating repo tags?

duffn:
1. I used version because it was simplest to ensure that we don’t delete the versioned docs when deploying but can look into it without: https://github.com/slack-ruby/slack-ruby-client/pull/450/files#diff-9cf2000c53760d837a449f874e53f792819108d3a4bf346336d0f7d082deae2cR50
2. We could enumerate though would need to ensure we only get 2.1.0 and up because previous versions are ignored since they won’t have docs. 

Let’s just keep the readme then. Though that was the reason I raised the issue before doing the work. The readme is definitely simpler to maintain. The site adds some “pizazz” but maybe not much more than that.

dblock:
I don't dislike the pizazz, but it sounds like you don't feel strongly about it either? 

In terms of pros we get easily linked old docs. That's a good thing. I think if we can automate the part where we update the stable release section to say "this is docs for version X" during build from tags, it would simplify releases, and therefore is quite valuable.

So what do you think about publishing the README without gutting it into new .md files into a new site? We would get old docs published by the same process by doing it once manually from tags.

duffn:
I'd feel more strongly if there docs were larger like https://docs.metasploit.com/. But for these docs, I don't have a strong preference. 

I'll take a look at the other ideas.

duffn:
I think now I'm more inclined to just suggest adding links to different versions in the README in lieu of creating the docs site. They can be generated during the release process and updated in the README. I can add this to the release doc. 


```bash
#!/bin/bash

echo "## Versions"
for tag in $(git tag --sort="-refname"); do
  echo "- [$tag](https://github.com/slack-ruby/slack-ruby-client/blob/$tag/README.md)"
done
```

dblock:
I am not sure we need a long list of versions. Latest feels good enough, no?

duffn:
Sure, one or two seem fine - just showing how they could be generated.

PR #452, opened 2023-03-21
Automate multi-argument requirements in Web API

jmanian:
I.e. methods that require exactly one of or at least one of a set of arguments.

It was not trivial to maintain the exact error messages that had been patched in previously, so the error messages have changed slightly. Otherwise there should be no functional change to this.

Also many auto-generated tests have changed due to slack-ruby/slack-api-ref#65.


spec/slack/web/api/endpoints/views_spec.rb
```diff
@@ -6,27 +6,60 @@
...
+      expect(client).to receive(:post).with('views.push', {interactivity_pointer: %q[12345.98765.abcd2358fdea], view: %q[]})
+      client.views_push(interactivity_pointer: %q[12345.98765.abcd2358fdea], view: %q[])
+
+      expect { client.views_push(trigger_id: %q[12345.98765.abcd2358fdea], interactivity_pointer: %q[12345.98765.abcd2358fdea], view: %q[]) }.to raise_error ArgumentError, /Exactly one of/
+    end
     it 'encodes view as json' do
-      expect(client).to receive(:post).with('views.push', {view: %q[{"data":["data"]}], trigger_id: %q[]})
-      client.views_push(view: {:data=>["data"]}, trigger_id: %q[])
+      expect(client).to receive(:post).with('views.push', {view: %q[{"data":["data"]}], trigger_id: %q[12345.98765.abcd2358fdea]})
+      client.views_push(view: {:data=>["data"]}, trigger_id: %q[12345.98765.abcd2358fdea])
```
dblock, line 45: Surprised rubocop is happy with this one.

dblock (approved):

PR #480, opened 2023-06-12
add common message formatting utilities

chrisbloom7:
I've been meaning to backport these slack message formatting utilities for a while. Finally getting around to it. I use them quite often and figured they may be useful to others as well.


dblock:
Looks good! Fix CHANGELOG and document in README?

Issue #482, opened 2023-06-22
Error details aren't passed through


kevinrobayna:
I see we have cassettes here https://github.com/slack-ruby/slack-ruby-client/blob/bf5bb4be3e03e9bd74d22942a61d5611f5c1d997/spec/fixtures/slack/web/views_open_error.yml#L79

but the error we get, if something fails, is just `invalid_arguments`, is this on purpose?

dblock:
Do you mean it doesn't include `response_metadata`? We should, it's not on purpose.

kevinrobayna:
Actually, the response metadata is present 😅 

I was expecting when printing e that I'd get all the details from the error (instead I was only getting the error message)

```

        client.views_open(
          view: view,
          trigger_id: request[:trigger_id]
        )
      rescue Slack::Web::Api::Errors::InvalidArguments => e
        puts e # prints invalid_arguments
        puts e.error # prints invalid_arguments
        puts e.response_metadata # prints the error from slack's api
      end
```

kevinrobayna:
Closing since people can get the error details as expected I was just being silly 😅

dblock:
We could change `to_s` to display it.

kevinrobayna:
yeah! raised a PR for it, let me know your thoughts

https://github.com/slack-ruby/slack-ruby-client/pull/483

PR #483, opened 2023-06-23
feat: Include details for web api errors

kevinrobayna:
fixes: https://github.com/slack-ruby/slack-ruby-client/issues/482

Currently when we get the error message we get no details of what went wrong unless we get the response_metadata for the error which is not intuitive as most people when they get an error and print the exception they would expect to see as much details as possible.

Additionally updated ruby-version in the project to be the latest as most people would be using that one as their default, though not changing the one required by the gem


lib/slack/web/api/errors/slack_error.rb
```diff
@@ -22,6 +22,12 @@ def errors
           def response_metadata
             response.body.response_metadata
           end
+
+          def to_s
```
kevinrobayna, line 26: the approach I took here is to put as many details as possible but without empty details i.e. the errors list would not be added if there are none, same as the response metadata.

.ruby-version
```diff
@@ -1 +1 @@
-2.7.6
+3.2.0
```
kevinrobayna, line 1: let me know if you want me to remove this from the PR 😅

dblock: Probably, but this is just the default version being used, it has little consequence on anything. We should probably just delete this file.

dblock (changes requested):
Add some specs with actual text? Let's see how it looks?

If we're going to start returning complex detailed errors we should adopt a better structured approach and even support localization. Check out https://github.com/dblock/ruby-enum/blob/master/lib/ruby-enum/errors/base.rb and how it constructs errors, I would copy everything from there.

Issue #485, opened 2023-06-26
Rate-limiting while trying to build RealTime::Stores::Store cache leads to keep_alive? failures


milestruecar:
https://github.com/slack-ruby/slack-ruby-client/blob/bf5bb4be3e03e9bd74d22942a61d5611f5c1d997/lib/slack/real_time/stores/store.rb#L328-L378

on initial connect, while cycling through users and conversations in this `hello` method, we're getting rate-limited and receiving 429 responses from slack. it looks like there's retry logic somewhere that's working, but somewhere along the way, before the cache actually completes (but never in a consistent spot), the connection just kinda dies, followed by another rtm.connect and another attempt to build the cache, which leads to more rate limiting, then another disconnect/reconnect, and so on forever.

after messing with the code locally, in this `keep_alive?` method https://github.com/slack-ruby/slack-ruby-client/blob/bf5bb4be3e03e9bd74d22942a61d5611f5c1d997/lib/slack/real_time/client.rb#L100-L116 while we're getting 429ed, the `@socket.time_since_last_message` eventually grows beyond `(websocket_ping * 2)` and knocks everything over. After adding some sleep calls to the cache build to try to mitigate the 429s and then disabling that keep_alive? entirely, the cache finished building and the app finally went into a steady state.

i had some ideas for possible fixes (longer timeout threshold if the last response was a 429 which i'm not entirely sure how to interrogate inside that `keep_alive?` method, just disabling `keep_alive?` until after the cache finishes which seems... not ideal, or adding sleeps to the user and conversation loops in `hello` method which would probably be silly for people with smaller slack footprints) but none of them seem great, and i'm not exactly a socket expert. so i wanted some guidance on what would make sense to y'all before i put together a PR

milestruecar:
while fiddling with this to try to find some short term solution, and noticed with a higher websocket_ping time set that even when we aren't getting rate-limited, `socket.time_since_last_message` just keeps going up. sample output from local branch with some extra logging thrown in
`[2023-06-26T16:05:30.778-07:00] INFO: POST https://slack.com/api/conversations.list
[2023-06-26T16:05:30.778-07:00] INFO: Accept: "application/json; charset=utf-8"
User-Agent: "Slack Ruby Client/2.1.1"
Authorization: [token]
Content-Type: "application/x-www-form-urlencoded"
[2023-06-26T16:05:30.864-07:00] INFO: @socket connected? true
[...]

Issue #237, opened 2018-11-16
Moving work off the main thread


kstole:
I haven't had time to contribute recently but I've been watching the Slack disconnects issues carefully as my bot is now disconnecting almost every night. I understand switching to async may be advantageous. One of the things I used heavily before was `Celluloid.defer` to prevent responding to commands blocking the bot from responding to other commands. This may be a question better suited for @ioquatix but is there an appropriate replacement for this using async?

dblock:
Generally I would like to include clear recommendation in README for this and possible default code that responds to commands into a defer-like behavior by default.

kstole:
Some discussion of this in https://github.com/socketry/async-websocket/issues/2

Seems like `Celluloid.defer` uses a separate thread to make the block asynchronous and `async` doesn't have an exact match. `Async.run do |task|` can be used and then `task.sleep` doesn't block and HTTP calls can be made asynchronous with `async/http/faraday` but not a simple cut/paste solution.

ioquatix:
> Seems like Celluloid.defer uses a separate thread to make the block asynchronous and async doesn't have an exact match.

I gave you an implementation for exactly that on the linked ticket :)

ioquatix:
By the way, using a thread/defer defeats the entire purpose of being asynchronous.

dblock:
I'd really appreciate if someone could clarify all this in the README, especially for async noobs like me. I just want to know what I am supposed to do :)

PR #486, opened 2023-06-27
Async handler support for Slack::RealTime::Config and Client

milestruecar:
Addresses https://github.com/slack-ruby/slack-ruby-client/issues/485


dblock:
LGTM, needs fixing tests, CHANGELOG, and maybe we can refactor the implementation a bit into private methods so that it's a bit more digestible?

dblock:
While you're at it, see if putting an `Async.run` in the `on` handler works for you? I think I'd prefer that change, as this problem exists for any long running task as a response to a Slack message. It would address https://github.com/slack-ruby/slack-ruby-client/issues/237.

milestruecar:
> While you're at it, see if putting an `Async.run` in any event (`on`) handler works for you? I think I'd prefer that change, as this problem exists for any long running task as a response to a Slack message. It would address #237.

are you talking about just wrapping this in `Async.run`? https://github.com/slack-ruby/slack-ruby-client/blob/bf5bb4be3e03e9bd74d22942a61d5611f5c1d997/lib/slack/real_time/client.rb#L34-L37

dblock:
> to
> ```
> handlers.each do |handler|
>   Async.run { store.instance_exec(data, self, &handler) }
> end
> ```
> 
> and it looks like it's working for my extremely narrow use case. Is that what you were after?

Right that's what I am after. 

Should it wrap each handler or all handlers? Is there a change in exception handling when one handler fails with the wrapped code?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * [#475](https://github.com/slack-ruby-client/pulls/475): Update API from [slack-api-ref@977dad5](https://github.com/slack-ruby/slack-api-ref/commit/977dad5) - [@slack-ruby-ci-bot](https://github.com/apps/slack-ruby-ci-bot).
 * [#476](https://github.com/slack-ruby-client/pulls/476): Update API from [slack-api-ref@d0b2989](https://github.com/slack-ruby/slack-api-ref/commit/d0b2989) - [@slack-ruby-ci-bot](https://github.com/apps/slack-ruby-ci-bot).
 * [#478](https://github.com/slack-ruby-client/pulls/478): Update API from [slack-api-ref@d797055](https://github.com/slack-ruby/slack-api-ref/commit/d797055) - [@slack-ruby-ci-bot](https://github.com/apps/slack-ruby-ci-bot).
+* [#486](https://github.com/slack-ruby/slack-ruby-client/pull/486): Async cache build for slack::realtime::stores::store - [@milestruecar](https://github.com/milestruecar).
```
dblock, line 14: Capitalize and quote `Slack::Realtime::...` to match actual class name.

dblock:
If this "just works", then I'd like a spec that ensures that handlers are run inside an async block, and we should talk about backwards compatibility and something in the README that says that handlers execute async. Then, what could go wrong? :) Or maybe this should be optional behind a configuration option?

milestruecar:
> > to
> > ```
> > handlers.each do |handler|
> >   Async.run { store.instance_exec(data, self, &handler) }
> > end
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and it looks like it's working for my extremely narrow use case. Is that what you were after?
> 
> Right that's what I am after.
> 
> Should it wrap each handler or all handlers? Is there a change in exception handling when one handler fails with the wrapped code?

i guess it depends on expected behavior, this makes it... more async? if order of handler execution matters, it would be better to have the entire block wrapped in a single async task. if order of execution doesn't matter (and everyone is fine spinning up some arbitrary number of async tasks) then this implementation seemed to work as expected for me

dblock:
Good point on order of execution. Current behavior is that it's predictable, so at a minimum we should keep the current behavior. How about introducing an option? We could call it `async_handlers` and it could take one of `:none` , `:all` and `:each`?

milestruecar:
tried implementing an async_handlers option with none all and each but i can't tell if all and each are actually doing anything meaningfully different. everything seems to fire off in the same order, the only real difference I can see is the each option is throwing some uncaught eof errors that i'm having trouble chasing down while crawling through the conversations.list loop

milestruecar:
just having :all as the default behavior makes sense to me, i can't picture a scenario in which someone would want to block the client from receiving from slack while the RealTime::Stores::Store cache gets hydrated, and i'm having trouble creating a scenario in which :each does anything meaningfully different

milestruecar:
you were asking about exceptions earlier, here's an example of one of those eof errors in an async task:
```
    1m     warn: Async::Task [oid=0x5d5c] [ec=0x5d70] [pid=8722] [2023-06-30 12:48:29 -0700]
               | Task may have ended with unhandled exception.
               |   Faraday::SSLError: SSL_connect returned=1 errno=0 peeraddr=54.70.179.16:443 state=error: unexpected eof while reading
               |   → [local path]/.rbenv/versions/3.2.1/lib/ruby/3.2.0/net/protocol.rb:46 in `connect_nonblock'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/3.2.0/net/protocol.rb:46 in `ssl_socket_connect'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/3.2.0/net/http.rb:1342 in `connect'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/3.2.0/net/http.rb:1248 in `do_start'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/3.2.0/net/http.rb:1237 in `start'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-net_http-3.0.2/lib/faraday/adapter/net_http.rb:112 in `request_with_wrapped_block'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-net_http-3.0.2/lib/faraday/adapter/net_http.rb:102 in `perform_request'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-net_http-3.0.2/lib/faraday/adapter/net_http.rb:66 in `block in call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/adapter.rb:45 in `connection'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-net_http-3.0.2/lib/faraday/adapter/net_http.rb:65 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/middleware.rb:17 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/response/logger.rb:23 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/middleware.rb:17 in `call'
               |     [local path]/slack-ruby-client/lib/slack/web/faraday/response/wrap_error.rb:16 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/middleware.rb:17 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/middleware.rb:17 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/middleware.rb:17 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/request/url_encoded.rb:25 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-multipart-1.0.4/lib/faraday/multipart/middleware.rb:28 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/rack_builder.rb:153 in `build_response'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/connection.rb:444 in `run_request'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/connection.rb:280 in `post'
               |     [local path]/slack-ruby-client/lib/slack/web/faraday/request.rb:25 in `request'
               |     [local path]/slack-ruby-client/lib/slack/web/faraday/request.rb:11 in `post'
               |     [local path]/slack-ruby-client/lib/slack/web/api/endpoints/conversations.rb:248 in `conversations_list'
               |     [local path]/slack-ruby-client/lib/slack/web/pagination/cursor.rb:25 in `block in each'
               |     [local path]/slack-ruby-client/lib/slack/web/pagination/cursor.rb:22 in `loop'
               |     [local path]/slack-ruby-client/lib/slack/web/pagination/cursor.rb:22 in `each'
               |     [local path]/slack-ruby-client/lib/slack/web/api/endpoints/conversations.rb:244 in `conversations_list'
               |     [local path]/slack-ruby-client/lib/slack/real_time/stores/store.rb:356 in `block in <class
               |     [local path]/slack-ruby-client/lib/slack/real_time/client.rb:240 in `instance_exec'
               |     [local path]/slack-ruby-client/lib/slack/real_time/client.rb:240 in `block (2 levels) in run_handlers'
               |     [local path]/slack-ruby-client/lib/slack/real_time/client.rb:239 in `each'
               |     [local path]/slack-ruby-client/lib/slack/real_time/client.rb:239 in `block in run_handlers'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/async-1.31.0/lib/async/task.rb:261 in `block in make_fiber'
               |   Caused by OpenSSL::SSL::SSLError: SSL_connect returned=1 errno=0 peeraddr=54.70.179.16:443 state=error: unexpected eof while reading
               |   → [local path]/.rbenv/versions/3.2.1/lib/ruby/3.2.0/net/protocol.rb:46 in `connect_nonblock'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/3.2.0/net/protocol.rb:46 in `ssl_socket_connect'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/3.2.0/net/http.rb:1342 in `connect'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/3.2.0/net/http.rb:1248 in `do_start'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/3.2.0/net/http.rb:1237 in `start'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-net_http-3.0.2/lib/faraday/adapter/net_http.rb:112 in `request_with_wrapped_block'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-net_http-3.0.2/lib/faraday/adapter/net_http.rb:102 in `perform_request'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-net_http-3.0.2/lib/faraday/adapter/net_http.rb:66 in `block in call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/adapter.rb:45 in `connection'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-net_http-3.0.2/lib/faraday/adapter/net_http.rb:65 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/middleware.rb:17 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/response/logger.rb:23 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/middleware.rb:17 in `call'
               |     [local path]/slack-ruby-client/lib/slack/web/faraday/response/wrap_error.rb:16 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/middleware.rb:17 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/middleware.rb:17 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/middleware.rb:17 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/request/url_encoded.rb:25 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-multipart-1.0.4/lib/faraday/multipart/middleware.rb:28 in `call'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/rack_builder.rb:153 in `build_response'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/connection.rb:444 in `run_request'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/faraday-2.7.9/lib/faraday/connection.rb:280 in `post'
               |     [local path]/slack-ruby-client/lib/slack/web/faraday/request.rb:25 in `request'
               |     [local path]/slack-ruby-client/lib/slack/web/faraday/request.rb:11 in `post'
               |     [local path]/slack-ruby-client/lib/slack/web/api/endpoints/conversations.rb:248 in `conversations_list'
               |     [local path]/slack-ruby-client/lib/slack/web/pagination/cursor.rb:25 in `block in each'
               |     [local path]/slack-ruby-client/lib/slack/web/pagination/cursor.rb:22 in `loop'
               |     [local path]/slack-ruby-client/lib/slack/web/pagination/cursor.rb:22 in `each'
               |     [local path]/slack-ruby-client/lib/slack/web/api/endpoints/conversations.rb:244 in `conversations_list'
               |     [local path]/slack-ruby-client/lib/slack/real_time/stores/store.rb:356 in `block in <class
               |     [local path]/slack-ruby-client/lib/slack/real_time/client.rb:240 in `instance_exec'
               |     [local path]/slack-ruby-client/lib/slack/real_time/client.rb:240 in `block (2 levels) in run_handlers'
               |     [local path]/slack-ruby-client/lib/slack/real_time/client.rb:239 in `each'
               |     [local path]/slack-ruby-client/lib/slack/real_time/client.rb:239 in `block in run_handlers'
               |     [local path]/.rbenv/versions/3.2.1/lib/ruby/gems/3.2.0/gems/async-1.31.0/lib/async/task.rb:261 in `block in make_fiber'
```

dblock:
> i can't picture a scenario in which someone would want to opt into :none and block the client from receiving from slack while the RealTime::Stores::Store cache gets hydrated, and i'm having trouble creating a scenario in which :each does anything meaningfully different

It's backwards incompatible, and would require a major version bump. I'm just trying to ask to default to the current behavior. Btw, not everybody uses stores. I disable that in my bots.

milestruecar:
first attempt at a :none and :all async_handlers config option. i don't love rescuing NoMethodError to catch invalid config options but couldn't figure out how to fail faster in the original configure call/yield

milestruecar:
there might be a cleaner way to implement this than a case statement. i tried a tricky metaprogramming thing first but it was pretty gross. idk what your preference is

lib/slack/real_time/client.rb
```diff
@@ -234,13 +234,24 @@ def dispatch(event)
 
       def run_handlers(type, data)
         handlers = store.class.events[type.to_s]
-        handlers.each do |handler|
-          store.instance_exec(data, self, &handler)
+        case @async_handlers
```
dblock, line 237: Use `case async_handlers` so it can be overwritten.

lib/slack/real_time/client.rb
```diff
@@ -234,13 +234,24 @@ def dispatch(event)
...
+        when :all 
+          Async.run { handlers_loop(handlers, data) }
+        when :none 
+          handlers_loop(handlers, data)
+        else
+          raise "Invalid Slack::RealTime::Client config! async_handlers must be :all or :none"
```
dblock, line 243: Let's specialize a `InvalidAsyncHandlersError` like we do for `NoConcurrencyError`, or raise a `RuntimeError` explicitly. I would include the current value, so "Invalid value "xyz" config#async_handlers, must be :all or :none." or something like that.

Add a period at the end since this is a full sentence.

dblock (changes requested):
Looks good, just needs tests and README.

README.md
```diff
@@ -432,6 +432,7 @@ websocket_proxy | Connect via proxy, include `:origin` and `:headers`.
 start_options   | Options to pass to `rtm.connect`, default is `{ request: { timeout: 180 } }`.
 store_class     | Local store class, default is an in-memory `Slack::RealTime::Stores::Starter`.
 store_options   | Options to initialize the store, default is `{}`.
+async_handlers  | Option to run handlers asynchronously. Valid options are `:all` or `:none`, default is `:none`
```
dblock, line 435: Add period at the end.

dblock (changes requested):
Tiny nit and this is good to go!

Fix rubocop with `rubocop -a ; rubocop --auto-gen-config`.

milestruecar:
do i need to somehow skip that test if CONCURRENCY isn't set? it seems like async isn't available with concurrency=none

dblock:
> do i need to somehow skip that test if CONCURRENCY isn't set? it seems like async isn't available with concurrency=none

I see what's happening. This comes from the fact that we used to support different concurrency libraries. So the async handler option is something that may have different implementations depending on the async library. Rather than muting the test add a method in `Slack::RealTime::Socket` called `run_async(&_block)` that raises `NotImplemented`, then override it in `Slack::RealTime::Concurrency::Async::Socket` with `Async.run ...` or something similar to `def start_async(client)`.

Then condition the test that actually runs async to the existence of the `CONCURRENCY` env.

milestruecar:
I'm not sure how to effectively test this since the socket is mocked

```
      context 'when config#async_handlers is :all', if: ENV['CONCURRENCY'] == 'async-websocket' do
          before do
            client.async_handlers = :all
            allow(socket).to receive(:run_async)
          end

          after do
            client.async_handlers = :none
          end

          it 'returns an Async::Task' do
            expect(client.send(:run_handlers, 'example', {})).to be_a Async::Task
          end
        end
```
this just returns `nil` for `run_async`

i know i can explicitly give a return `allow(socket).to receive(:run_async).and_return(Async::Task)` or something like that but then i'm not actually testing that method

spec/slack/real_time/client_spec.rb
```diff
@@ -159,16 +159,31 @@
...
 
       describe '#run_handlers' do
-        context 'when store has no event hooks' do
+        before do
+          @events = client.store.class.events.dup
+          client.store.class.events.clear
+        end
+
+        after do
+          client.store.class.events.merge!(@events)
```
dblock, line 168: I know this was copied, but should this be an assignment? Otherwise you inherit events added during the test.

dblock (changes requested):

spec/slack/real_time/client_spec.rb
```diff
@@ -159,16 +159,31 @@
...
+        end
+
+        context 'when config#async_handlers is :all', if: ENV['CONCURRENCY'] == 'async-websocket' do
           before do
-            @events = client.store.class.events.dup
-            client.store.class.events.clear
+            client.async_handlers = :all
+            allow(socket).to receive(:run_async).and_return(Async.run { nil })
```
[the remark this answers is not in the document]
milestruecar: that first and last points are what i was talking about in my last comment, it doesn't return an `Async::Task` out of the box because the socket is mocked. It just gets `nil` back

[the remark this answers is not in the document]
milestruecar: when i run without that explicit return i get this
```
Failures:

  1) Slack::RealTime::Client client with a full store #start! #run_handlers when config#async_handlers is :all returns an Async::Task
     Failure/Error: expect(client.send(:run_handlers, 'example', {})).to be_a Async::Task
       expected nil to be a kind of Async::Task
     # ./spec/slack/real_time/client_spec.rb:182:in `block (6 levels) in <top (required)>'
```

spec/slack/real_time/client_spec.rb
milestruecar: that was just lifted from the prior code but i can change it. i have no idea what the expected performance is, i just re-used what was there.

spec/slack/real_time/client_spec.rb
```diff
@@ -159,16 +159,31 @@
...
-            client.store.class.events.merge!(@events)
+            client.async_handlers = :none
           end
 
+          it 'returns an Async::Task' do
+            expect(client.send(:run_handlers, 'example', {})).to be_a Async::Task
+          end
+        end
+
+        context 'when store has no event hooks' do
```
milestruecar, line 186: you get an implicit test of the :none behavior here, since this test returns an empty array only when async_handlers is :none

spec/slack/real_time/client_spec.rb
milestruecar: making this an assignment instead of a merge broke... a bunch of other tests. so i guess this is necessary

dblock: Yep, this was written because `events=` doesn't exist. It could be `client.store.class.events.instance_variable_set(:@events, ...)`, but a cleaner fix would be to implement a `reset` method on the store just like for config. Either way that's outside of the scope of this PR.

spec/slack/real_time/client_spec.rb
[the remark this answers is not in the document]
dblock: This was expected. In this test the socket client (the part that is actually async) is mocked rather than using the real concurrency class.

```ruby
let(:ws) { double(Slack::RealTime::Concurrency::Mock::WebSocket, on: true) }
```

It therefore never invokes the real `Slack::RealTime::Concurrency::Async::Socket` and so you can't check the return of that.

The test here is good, it checks that 1) when `async_handlers=:all` we call `run_async` on the socket class, and 2) when `async_handlers=:none` we don't call it. If we want to check that `run_async` works as expected for the real `Slack::RealTime::Concurrency::Async::Socket` we take it out into a separate spec for that, which I added in https://github.com/milestruecar/slack-ruby-client/pull/1. I think we need to add tests there (in a future PR), which reminds me to do code coverage ;)

milestruecar:
> Looks like some silliness in requires after my PR.
> 
> ```
> 
> LoadError:
>   cannot load such file -- async/websocket
> # ./lib/slack/real_time/concurrency/async.rb:2:in `require'
> # ./lib/slack/real_time/concurrency/async.rb:2:in `<top (required)>'
> # ./spec/slack/real_time/concurrency/clients/async_spec.rb:4:in `require'
> # ./spec/slack/real_time/concurrency/clients/async_spec.rb:4:in `<top (required)>'
> ```
> 
> LMK if you need help finishing this!

it's due to the new test file
```
RSpec.describe Slack::RealTime::Concurrency::Async::Socket, if: ENV['CONCURRENCY'] == 'async-websocket' do
```
i think it's trying to include that class regardless of ENV['CONCURRENCY'] on initialization, i'm not sure how to get around this, short of just fully excluding that file on rspec command line

dblock:
This is because we use `Slack::RealTime::Concurrency::Async::Socket`, and it auto-resolves `async`, requires it, then fails because the async-websocket library is not included in the Gemfile when `CONCURRENCY` is not set. To fix we change that to a string (so it doesn't resolve the class name), and condition it similarly to the other tests that skip with different values of CONCURRENCY for consistently mostly.

https://github.com/milestruecar/slack-ruby-client/pull/2

milestruecar:
> This is because we use `Slack::RealTime::Concurrency::Async::Socket`, and it auto-resolves `async`, requires it, then fails because the async-websocket library is not included in the Gemfile when `CONCURRENCY` is not set. To fix we change that to a string (so it doesn't resolve the class name), and condition it similarly to the other tests that skip with different values of CONCURRENCY for consistently mostly.
> 
> [milestruecar#2](https://github.com/milestruecar/slack-ruby-client/pull/2)

i... did not know you could just drop a string in there. obviously i only rspec at about a 3rd grade level. Thanks!

dblock (approved):

dblock:
Merged, thank you!

Issue #388, opened 2021-12-02
chat.unfurl allows unfurl_id and source instead of channel and ts


MasterOdin:
See https://api.slack.com/changelog/2021-08-changes-to-unfurls.

The chat.unfurl endpoint now allows for using either `unfurl_id` and `source` _or_ `channel` and `ts`.

dblock:
Please contribute!

kwent:
This appears to be fixed in v3.0.0 — the `chat_unfurl` method no longer raises `ArgumentError` for missing `:channel` when `unfurl_id` + `source` are provided instead.

The fix came via the API ref update in commit 9fff2eb (slack-api-ref@c408935, 2025-09-21).

For anyone on v2.x, you can work around it by calling the Slack API directly via Faraday:

```ruby
params = {token: integration.token, unfurl_id: unfurl_id, source: source}
params[:unfurls] = unfurls.to_json
params[:metadata] = metadata.to_json if metadata.present?
response = Faraday.new(url: "https://slack.com/api").post("chat.unfurl", params)
```

Would be good to close this issue since the fix is released.

dblock:
Thanks for confirming the fix @kwent!

PR #502, opened 2023-11-01
feat: Allows either unfurl_id & source or channel & ts in `chat.unfurl` method

muhsin-k:
Fixes https://github.com/slack-ruby/slack-ruby-client/issues/388

### DIFF
--- a/lib/slack/web/api/endpoints/chat.rb
+++ b/lib/slack/web/api/endpoints/chat.rb
@@ -236,8 +236,11 @@ module Slack
           # @see https://api.slack.com/methods/chat.unfurl
           # @see https://github.com/slack-ruby/slack-api-ref/blob/master/methods/chat/chat.unfurl.json
           def chat_unfurl(options = {})
-            raise ArgumentError, 'Required arguments :channel missing' if options[:channel].nil?
-            raise ArgumentError, 'Required arguments :ts missing' if options[:ts].nil?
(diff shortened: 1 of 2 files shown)


lib/slack/web/api/endpoints/chat.rb
```diff
@@ -236,8 +236,11 @@ def chat_scheduleMessage(options = {})
           # @see https://api.slack.com/methods/chat.unfurl
           # @see https://github.com/slack-ruby/slack-api-ref/blob/master/methods/chat/chat.unfurl.json
           def chat_unfurl(options = {})
-            raise ArgumentError, 'Required arguments :channel missing' if options[:channel].nil?
-            raise ArgumentError, 'Required arguments :ts missing' if options[:ts].nil?
+            if options[:channel].nil? || options[:ts].nil?
+              if options[:unfurl_id].nil? || options[:source].nil?
+                raise ArgumentError, 'Either a combination of :channel and :ts or :unfurl_id and :source is required'
+              end
+            end
```
dblock, line 243: This code is generated, so you'll have to either make a patch in https://github.com/slack-ruby/slack-ruby-client/tree/master/lib/slack/web/api/patches, or figure out a way to specify mutually exclusive options.

dblock: Note the code as is would allow `channel` and `source`, so you will want to do something else.

PR #507, opened 2024-01-30
Update README.md

gjtorikian:
`Faraday::UploadIO` has been [deprecated for some time](https://github.com/lostisland/faraday/blob/87e655f306454b49e459ac0a06d617cbad497fb4/CHANGELOG.md?plain=1#L45); this PR simply updates the README example to use the new class.


dblock (approved):
@greggroth Thanks. Care to add a test that uses this class? This way we'd catch the deprecation next time.

PR #516, opened 2024-07-03
CI: Add Ruby 3.3 to matrix

olleolleolle:
This PR adds Ruby 3.3 to the build matrix in the test suite.

Ruby 2.7 and Ruby 3.0 are both EOL, now, so their removal could be considered.


dblock (changes requested):
Thanks! Add a CHANGELOG entry that says "Add support for Ruby 3.3."

dblock (approved):

PR #517, opened 2024-07-03
CI: Try to run bundle install just once

olleolleolle:
This PR moves the environment variable-setting Step to before the "setup Ruby" step.

The Setup Ruby step does a `bundle install`.

That means that we can remove the explicit `bundle install` later on the file.


dblock:
Looks good, thanks.

PR #520, opened 2024-07-05
Support basic Markdown formatting

### LINKED ISSUE #249: Parse standards-formatted text into slack-specific mishmash
A casual google reveals at least four javascript libraries that offer parsing from standards-based text (markdown, html) into whatever the heck it is that slack accepts (some mishmash of markdown and angle-brackets links, • characters instead of lists, etc) [[1](https://github.com/mrq-cz/slackify-html), [2](https://github.com/jsarafajr/slackify-markdown), [3](https://github.com/integrations/html-to-mrkdwn), [4](https://github.com/omarish/pandoc-md-to-slack)].

This gem (or some other gem in this org?) should offer similar, officially-support functionality, in order to make it trivially simple to send html or markdown to the Slack API.

nbgoodall:
This PR adds basic support for Slack's markdown formatting, specifically **bold**, ~~strikethrough~~ and [links](https://example.com). Non-supported features like headings and tables are left as-is.

Closes #249.

### DIFF
--- a/lib/slack/messages/formatting.rb
+++ b/lib/slack/messages/formatting.rb
@@ -74,14 +74,10 @@ module Slack
         # @see https://api.slack.com/reference/surfaces/formatting#basic-formatting
         #
         def markdown(text)
-          # convert bold
-          slack_md = text.gsub(/\*\*(.*?)\*\*/, '*\1*')
-
-          # convert strikethrough
-          slack_md = slack_md.gsub(/~~(.*?)~~/, '~\1~')
-
-          # convert links
-          slack_md.gsub(/\[(.*?)\]\((.*?)\)/, '<\2|\1>')
+          text
+            .gsub(/\*\*(.*?)\*\*/, '*\1*') # bold
+            .gsub(/~~(.*?)~~/, '~\1~') # strikethrough
+            .gsub(/\[(.*?)\]\((.*?)\)/, '<\2|\1>') # links
         end
       end
     end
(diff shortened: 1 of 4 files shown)


lib/slack/messages/formatting.rb
```diff
@@ -68,6 +68,21 @@ def user_link(user_id)
...
+        def markdown(text)
+          # convert bold
+          slack_md = text.gsub(/\*\*(.*?)\*\*/, '*\1*')
+
+          # convert strikethrough
+          slack_md = slack_md.gsub(/~~(.*?)~~/, '~\1~')
+
+          # convert links
+          slack_md.gsub(/\[(.*?)\]\((.*?)\)/, '<\2|\1>')
+        end
```
dblock, line 85: Let’s chain them as `text.gsub…gsub…` to avoid the temporary variable.

dblock (changes requested):
Question: are we going to run into situations where the order of these replacements will matter? Like `**bold_italic_**bold` - meaning I wonder whether it’s worth implementing something that actually reads tokens sequentially.

Either way I’m good with a first cut like this. Nitpicks follow.

nbgoodall:
> Question: are we going to run into situations where the order of these replacements will matter? Like `**bold_italic_**bold` - meaning I wonder whether it’s worth implementing something that actually reads tokens sequentially.

@dblock I won't say never, although I haven't found any glaring cases. It wouldn't have converted `*italic*`, which I've now fixed (also accounting for `***bold and italic***` to `*_bold and italic_*`). If other issues come up I'll happily address them, and I've added another spec for nested formatting.

dblock (approved):

dblock:
Merged, thank you!

PR #525, opened 2024-08-02
Exclude spec files from gem package

amatsuda:
Here's a patch that excludes files under spec directory from the gem package.

With this patch, the gem package size shrinks as follows:
```
before: 355840 bytes
after:  293376 bytes
```


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ### 2.4.1 (Next)
 
-* Your contribution here.
```
dblock, line 3: Put this back please.

dblock (changes requested):
Looks good, see below.

CHANGELOG.md
amatsuda: Oops, sorry. Fixed!

dblock (approved):

Issue #514, opened 2024-06-17
Remove files.upload in favor of files.getUploadURLExternal, etc.


gregorbg:
The current README file (as of opening this issue) [references](https://github.com/slack-ruby/slack-ruby-client?tab=readme-ov-file#upload-a-file) the `client.files_upload` method to upload files. Under the hood, this uses the `files.upload` endpoint which has been deprecated by Slack: https://api.slack.com/changelog/2024-04-a-better-way-to-upload-files-is-here-to-stay

The article that I linked above describes the migration path, but it only has code samples for JS, Python and Java.
What is the recommended way to use the `files.getUploadURLExternal` endpoint with this client?

In particular, I can see that the response of the `files.getUploadURLExternal` invocation contains a link which Slack expects me (the client) to make a POST request to. That means I'd have to open and maintain my own HTTP client (Faraday etc.) **on top of** this library's Slack client. That seems a bit overkill, is there a better way?

dblock:
Sounds like we need to remove `files_upload` (maybe be coming in via #509 already) and reimplement something similar to Python's `files_upload_v2`. Want to give it a shot @gregorbg?

booleanbetrayal:
We just hit this issue after having to regenerate new "Modern Apps" (via a Legacy Bot) as part of a [deprecation notice](https://api.slack.com/changelog/2024-09-legacy-custom-bots-classic-apps-deprecation) remediation and now are unable to use the `files_upload` feature without `Slack::Web::Api::Errors::MethodDeprecated` errors, since any [new bots do not have access](https://api.slack.com/changelog/2024-04-a-better-way-to-upload-files-is-here-to-stay#change).

monkeyWzr:
Hi there @dblock , I'd love to give this one a shot!

dblock:
@monkeyWzr Don't let me stop you.

sny-tanaka:
Hello, there.
Is anyone working on this issue?

PR #532, opened 2025-01-30
Implement `files_upload_v2` helper to wrap new file upload sequential flow

gregsaab:
Fixes #514 

This adds a `Slack::Web::Api:Helpers::Files` module with a `files_upload_v2` helper method that implements [the new file upload flow](https://forums.slackcommunity.com/s/question/0D5Hq00009pjbwTKAQ/since-filesupload-is-being-deprecated-is-there-another-way-to-upload-files-to-slack-channels?language=en_US).

Two changes were made to the Faraday config/handlers:
- allow sending string content
- allow receiving string responses

Both of these are relevant to the `POST` to send the file contents (e.g. step 2 in the aforementioned new upload flow) as the `content-type` on the request and response are both non-json strings.

PR #533, opened 2025-01-31
Implement `files_upload_v2` helper to wrap new file upload sequential flow

gregsaab:
This is a replacement for PR #532 -- needed to create a new PR so that the head repo was on a feature branch.

Fixes https://github.com/slack-ruby/slack-ruby-client/issues/514

This adds a Slack::Web::Api:Helpers::Files module with a files_upload_v2 helper method that implements [the new file upload flow](https://forums.slackcommunity.com/s/question/0D5Hq00009pjbwTKAQ/since-filesupload-is-being-deprecated-is-there-another-way-to-upload-files-to-slack-channels?language=en_US).

Two changes were made to the Faraday config/handlers:
- allow sending string content
- allow receiving string responses

Both of these are relevant to the `POST` to send the file contents (e.g. step 2 in the aforementioned new upload flow) as the `content-type` on the request and response are both non-json strings.


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * [#527](https://github.com/slack-ruby/slack-ruby-client/pull/527): Explicitly require `racc` and `ostruct` - [@dblock](https://github.com/dblock).
 * [#528](https://github.com/slack-ruby/slack-ruby-client/pull/528): Don't treat asterisks in list items as italic in markdown - [@rspeicher](https://github.com/rspeicher).
 * [#530](https://github.com/slack-ruby/slack-ruby-client/pull/530): Specify dependency on logger - [@rwstauner](https://github.com/rwstauner).
+* [#533](https://github.com/slack-ruby/slack-ruby-client/pull/533): Implement new file upload helper - [@gregsaab](https://github.com/gregsaab).
```
dblock, line 7:
```suggestion
* [#533](https://github.com/slack-ruby/slack-ruby-client/pull/533): Added `files_upload_v2` - [@gregsaab](https://github.com/gregsaab).
```

README.md
```diff
@@ -171,7 +171,26 @@ general_channel = channels.detect { |c| c.name == 'general' }
...
-Upload a file with [files_upload](https://api.slack.com/methods/files.upload).
+Upload files with [sequenced API calls](https://api.slack.com/messaging/files#uploading_files).
+
+This library provides a helper method `files_upload_v2` that wraps the three separate API calls.
+
+```ruby
+contents = File.read('/users/me/results.pdf')
+client.files_upload_v2(
+  # required options
+  channels: 'C000000,C000001', # comma delimited channel ids, only one channel is required
```
dblock, line 182: Does the helper support arrays of channels? We transform these in the generated API calls. Check/add tests.

README.md
```diff
@@ -171,7 +171,26 @@ general_channel = channels.detect { |c| c.name == 'general' }
...
+  filename: 'results.pdf', # this is used for the file title, unless a :title option is provided
+  contents: contents, # the string contents of the file
```
dblock, line 184: Nit: we don't reuse the variable `contents`, so don't make one.

```suggestion
  contents: File.read('results.pdf'), # the string contents of the file
```

lib/slack/web/faraday/response/raise_error.rb
```diff
@@ -4,14 +4,39 @@ module Web
...
-          def on_complete(env)
+          def raise_if_too_many_requests!(env)
             raise Slack::Web::Api::Errors::TooManyRequestsError, env.response if env.status == 429
+          end
+
+          def raise_if_response_is_invalid_json!(env)
+            return if response_content_type_is_a_string?(env) || env.body.is_a?(Hash)
+
+            begin
+              raise JSON::ParserError
```
dblock, line 15: This looks odd. When would `ParserError` fail to raise? We should be just `raise ::Faraday::ParsingError.new(e, env.response)` here shouldn't we?

lib/slack/web/faraday/response/raise_error.rb
```diff
@@ -4,14 +4,39 @@ module Web
...
+            rescue StandardError => e
+              raise ::Faraday::ParsingError.new(e, env.response)
+            end
+          end
+
+          def response_content_type_is_a_string?(env)
+            env.response&.headers&.[]('content-type')&.include?('text/plain') || false
```
dblock, line 22: Content-type always starts with `text/plain`, right? Better than an `include?`. But ideally we should be parsing the mime type here and having a strong check. Maybe Faraday does this for us or already brings in a lib that can do it? Not sure it's worth it.

lib/slack/web/faraday/response/raise_error.rb
```diff
@@ -4,14 +4,39 @@ module Web
...
+          end
+
+          def on_complete(env)
+            raise_if_too_many_requests!(env)
+            raise_if_response_is_invalid_json!(env)
 
             body = env.body
-            return unless body
-            return if body['ok']
+            return if should_return?(env)
```
dblock, line 39: I think this is a behavior change we don't want. Before a JSON Slack API always assumed it was JSON and checked `ok`. Now if the sever returns plain text on an error (e.g. a proxy dumps an HTML page with content-type text/plain) we just pass this on.

The issue is that we are trying to workaround a response to a `post(upload_url, content)` with the same Faraday setup. Split https://github.com/slack-ruby/slack-ruby-client/blob/master/lib/slack/web/faraday/connection.rb into a base class with the proxy/SSL setup with all this special Slack middleware and another without, or create a new one without JSON processing and use that to do the data `post`?

dblock (changes requested):
I think this introduces a breaking behavior in the way errors in all the other JSON APIs are handled :( See below, we can make it simpler.

lib/slack/web/faraday/response/raise_error.rb
gregsaab: Yeah, I was considering making another connection, hut decided against. I'll go that route.

README.md
gregsaab: https://github.com/slack-ruby/slack-ruby-client/pull/533/commits/206f8b3b5ec7f3adef7023c9a080472541160246
👍

lib/slack/web/faraday/response/raise_error.rb
gregsaab: https://github.com/slack-ruby/slack-ruby-client/pull/533/commits/6792106161ebf0145c989f29b87f5caf537f2574

lib/slack/web/faraday/response/raise_error.rb
gregsaab: Removed this logic 👍

lib/slack/web/faraday/response/raise_error.rb
gregsaab: Removed logic 👍

CHANGELOG.md
gregsaab: https://github.com/slack-ruby/slack-ruby-client/pull/533/commits/fedd8add873ce96790013748876d590a4146f735

gregsaab:
I have seen a failure from this build step in the original PR for missing the `changelog.md` update, but I'm thinking this failure is not related to my change?
image

lib/slack/web/faraday/connection.rb
```diff
@@ -3,36 +3,44 @@ module Slack
   module Web
     module Faraday
       module Connection
-        private
+        def default_options
```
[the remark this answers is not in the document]
dblock: Run

```
rubocop -a
rubocop --auto-gen-config
```

and don't worry about it.

dblock (changes requested):
The Slack APIs are `*UploadExternal`, why don't we call `files_upload_v2` as `files_upload_external` as Slack intended it?

There's still a lot of `if`s going on inside `connection.rb` and flags like `expect_json_response`. This can be solved in a much cleaner way with OO by creating `BaseConnection` and extracting the common parts into it. Let me know if you want help with this.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * [#527](https://github.com/slack-ruby/slack-ruby-client/pull/527): Explicitly require `racc` and `ostruct` - [@dblock](https://github.com/dblock).
 * [#528](https://github.com/slack-ruby/slack-ruby-client/pull/528): Don't treat asterisks in list items as italic in markdown - [@rspeicher](https://github.com/rspeicher).
 * [#530](https://github.com/slack-ruby/slack-ruby-client/pull/530): Specify dependency on logger - [@rwstauner](https://github.com/rwstauner).
+* [#533](https://github.com/slack-ruby/slack-ruby-client/pull/533): Added `files_upload_v2`- [@gregsaab](https://github.com/gregsaab).
```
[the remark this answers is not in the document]
gregsaab: https://github.com/slack-ruby/slack-ruby-client/pull/533/commits/6c2e12ec8fd9909caa9525e678fd112d29408b90

gregsaab:
> The Slack APIs are `*UploadExternal`, why don't we call `files_upload_v2` as `files_upload_external` as Slack intended it?

To me, `v2` is more useful as it differentiates this implementation from the deprecated one. File uploads by definition are "external"

> There's still a lot of ifs going on inside connection.rb and flags like expect_json_response. This can be solved in a much cleaner way with OO by creating BaseConnection and extracting the common parts into it. Let me know if you want help with this.

To me, adding a few "if"s is following the pattern that seems to be in place in this file. I'm not seeing the value in adding multiple base classes of `Faraday::Connection`. Here is the original:
```
 def connection
          @connection ||=
            begin
              options = {
                headers: { 'Accept' => 'application/json; charset=utf-8' }
              }

              options[:headers]['User-Agent'] = user_agent if user_agent ##
              options[:proxy] = proxy if proxy ##
              options[:ssl] = { ca_path: ca_path, ca_file: ca_file } if ca_path || ca_file ##

              request_options = {}
              request_options[:timeout] = timeout if timeout ##
              request_options[:open_timeout] = open_timeout if open_timeout ##
              options[:request] = request_options if request_options.any?  ##

              ::Faraday::Connection.new(endpoint, options) do |connection|
                connection.request :multipart
                connection.request :url_encoded
                connection.use ::Slack::Web::Faraday::Response::RaiseError
                connection.response :mashify, mash_class: Slack::Messages::Message
                connection.response :json, content_type: /\b*$/
                connection.use ::Slack::Web::Faraday::Response::WrapError
                connection.response :logger, logger if logger. ##
                connection.adapter adapter
              end
            end
        end
```
So there are 7 lines that have conditionals on them and this PR adds 2 more IFs. How about I add an if block on `expect_json_response` so that only 1 more conditional is added?

dblock (changes requested):
I tried to simplify the approach in https://github.com/gregsaab/slack-ruby-client/pull/1, lmk what you think?

gregsaab:
@dblock thank you for pushing your suggestion!

dblock (approved):

dblock:
@gregsaab I re-read https://api.slack.com/changelog/2024-04-a-better-way-to-upload-files-is-here-to-stay and it looks like other client libraries all called this API files_upload_v2? If you still feel strongly about renaming `files_upload_external` to `files_upload_v2` I would take that PR before we release v. next. WDYT?

dblock:
@gregsaab I also opened https://github.com/slack-ruby/slack-ruby-client/issues/534 and https://github.com/slack-ruby/slack-ruby-client/issues/535, maybe you're interested in picking either up.

dblock:
2.5.0 has been released, much thanks!

https://x.com/dblockdotorg/status/1888613729077493893

Issue #539, opened 2025-02-08
Add support for channel_id in files_upload_v2.


dblock:
The documentation in https://api.slack.com/methods/files.completeUploadExternal claims that you can pass both `channels` and `channel_id` or none at all.

It's unlikely an accident, and Slack can potentially treat the file differently depending on whether one passed `channel_id` or `channels`. We should probably support `channel_id` being passed through in `files_uploadExternal` without mixing it with `channels`.

The API also supports not passing a channel, making the uploaded file private. 

cc: @gregsaab

PR #540, opened 2025-02-08
Added support for channel and channel_id, do not require channel.

dblock:
Closes https://github.com/slack-ruby/slack-ruby-client/issues/539.

cc: @gregsaab


lib/slack/web/api/helpers/files.rb
```diff
@@ -28,13 +32,26 @@ module Files
...
+            raise ArgumentError, 'Only one of :channel, :channels, or :channel_id is required' if channel_params.size > 1
+
+            complete_upload_request_params = params.slice(:initial_comment, :thread_ts)
+
+            if params[:channels]
+              complete_upload_request_params[:channels] = Array(params[:channels]).map do |channel|
+                conversations_id(channel: channel)['channel']['id']
+              end.uniq.join(',')
+            elsif params[:channel]
+              complete_upload_request_params[:channel] = conversations_id(
```
saeyoung-kim0, line 49: should this be :channel_id instead of :channel? I don't see channel as one of the available arguments in https://api.slack.com/methods/files.completeUploadExternal

dblock: Fixed in https://github.com/slack-ruby/slack-ruby-client/pull/548, releasing as 2.5.2.

PR #545, opened 2025-02-13
fix: `files_upload_v2` usage in README

LesnyRumcajs:
Fixed according to the current API

https://github.com/slack-ruby/slack-ruby-client/blob/b05610b0d70bed4ecb079260892478077737d313/lib/slack/web/api/helpers/files.rb#L34-L37


dblock:
You are right!

Issue #512, opened 2024-03-27
Slack::Event::Request#verify! should use secure_compare


mintuhouse:
https://github.com/slack-ruby/slack-ruby-client/blob/9e6d1171aecfb89d11ce55fb010cc3370666cd01/lib/slack/events/request.rb#L62

if using active_support something like `ActiveSupport::SecurityUtils.secure_compare`

dblock:
Good point, want to PR something along those lines @mintuhouse?

PR #553, opened 2025-04-06
Use secure_compare during signature verification

hieuk09:
Close https://github.com/slack-ruby/slack-ruby-client/issues/512

I decide to implement the method in the class directly because it's pretty simple. That way, we can avoid adding new dependency to the gem.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ### 2.5.3 (Next)
 
 * [#549](https://github.com/slack-ruby/slack-ruby-client/pull/549): Add group ID formatting support for message mentions - [@n0h0](https://github.com/n0h0).
+* [#553](https://github.com/slack-ruby/slack-ruby-client/pull/553): Use secure_compare during signature verification - [@hieuk09](https://github.com/hieuk09).
 * Your contribution here.
 
 ### 2.5.2 (2025/02/19)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ### 2.5.3 (Next)
 
 * [#549](https://github.com/slack-ruby/slack-ruby-client/pull/549): Add group ID formatting support for message mentions - [@n0h0](https://github.com/n0h0).
-* [#553](https://github.com/slack-ruby/slack-ruby-client/pull/553): Use secure_compare during signature verification - [@hieuk09](https://github.com/hieuk09).
+* [#553](https://github.com/slack-ruby/slack-ruby-client/pull/553): Use `secure_compare` during signature verification - [@hieuk09](https://github.com/hieuk09).
 * Your contribution here.
 
 ### 2.5.2 (2025/02/19)
--- a/lib/slack-ruby-client.rb
+++ b/lib/slack-ruby-client.rb
@@ -22,7 +22,7 @@ end
 begin
   require 'openssl'
 rescue LoadError # rubocop:disable Lint/SuppressedException
-  # Used in slack/web/config
+  # Used in slack/web/config and slack/utils/security
 end
 require_relative 'slack/web/config'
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ### 2.5.3 (Next)
 
 * [#549](https://github.com/slack-ruby/slack-ruby-client/pull/549): Add group ID formatting support for message mentions - [@n0h0](https://github.com/n0h0).
+* [#553](https://github.com/slack-ruby/slack-ruby-client/pull/553): Use secure_compare during signature verification - [@hieuk09](https://github.com/hieuk09).
```
dblock, line 4:
```suggestion
* [#553](https://github.com/slack-ruby/slack-ruby-client/pull/553): Use `secure_compare` during signature verification - [@hieuk09](https://github.com/hieuk09).
```

dblock (changes requested):
Thanks! Isn't this available in ActiveSupport that we already depend on?

https://api.rubyonrails.org/classes/ActiveSupport/SecurityUtils.html#method-c-secure_compare

hieuk09:
I think `activesupport` is currently a test-only dependency. I could not find it in `slack-ruby-client.gemspec` 🤔 .

dblock:
> I think `activesupport` is currently a test-only dependency. I could not find it in `slack-ruby-client.gemspec` 🤔 .

Yes, you're right, my bad.

dblock:
We need to wait till coveralls comes back up hopefully soon for CI to pass ...

dblock:
@hieuk09 I am pretty sure we have `OpenSSL` defined as. dependency of Faraday, so should we use that implementation?

See https://github.com/rails/rails/blob/3235827585d87661942c91bc81f64f56d710f0b2/activesupport/lib/active_support/security_utils.rb#L10 for how ActiveSupport does it.

hieuk09:
Thanks for the info. Let me try that out.

hieuk09:
Using openssl works, thanks for the catch 👍 

I see that this method is available from openssl 2.2.0. Should I make that dependency explicit in gemspec?

dblock:
> I see that this method is available from openssl 2.2.0. Should I make that dependency explicit in gemspec?

OpenSSL is part of the standard library of Ruby (a [default gem](https://docs.ruby-lang.org/en/master/standard_library_rdoc.html)), but Ruby itself can work with openssl 1.x as well per https://github.com/ruby/openssl. I could be wrong, I think this means that saying `>= 2` in .gemspec won't do what we want, the version of the gem will be >= 2, but the version of OpenSSL may still be 1.x and that call would fail.

I think this is why ActiveSupport does `if defined?(OpenSSL.fixed_length_secure_compare)` with a fallback in [the code](https://github.com/rails/rails/blob/3235827585d87661942c91bc81f64f56d710f0b2/activesupport/lib/active_support/security_utils.rb#L10), I an afraid we have to copy that entire implementation. You should copy the security_utils.rb file entirely into this project + its tests (or write new ones), reference where they came from in the header, and namespace them under `Slack::` to avoid a conflict. 

Sorry for the back and forth and thanks for hanging in here with me!

hieuk09:
Thanks for the explanation, I learn a lot from the back-and-forth so it's great :D. Let me check and push another fix.

dblock (commented):
Good! Let's put the `require`'s into https://github.com/slack-ruby/slack-ruby-client/blob/master/lib/slack-ruby-client.rb just to be consistent?

CI caught test failures with Ruby 2.x, take a look, too.

https://github.com/slack-ruby/slack-ruby-client/actions/runs/14307965727/job/40096005485?pr=553

hieuk09:
Thanks, the tests are passing now 😄

dblock (approved):

dblock:
Good work, thanks for hanging in here with me @hieuk09!

PR #555, opened 2025-04-29
More efficient name-to-id translation

eizengan:
The `conversations_id` method paginates with default options over [conversations.list](https://api.slack.com/methods/conversations.list) responses to find a conversation with a particular name. The `conversations.list` endpoint has a [Tier 2 rate limit](https://api.slack.com/apis/rate-limits#tier_t2), which - when combined with the low default page size - can cause just a single call in a workspace with 2000+ channels to result in rate limiting. This change enables use of non-default page sizes within `conversations_id` (and `id_for`, more generally), thus reducing the risk of rate limiting within such workspaces.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 * [#549](https://github.com/slack-ruby/slack-ruby-client/pull/549): Add group ID formatting support for message mentions - [@n0h0](https://github.com/n0h0).
 * [#553](https://github.com/slack-ruby/slack-ruby-client/pull/553): Use `secure_compare` during signature verification - [@hieuk09](https://github.com/hieuk09).
-* [#555](https://github.com/slack-ruby/slack-ruby-client/pull/555): Make page size in resolving conversation IDs configurable - [@eizengan](https://github.com/eizengan).
+* [#555](https://github.com/slack-ruby/slack-ruby-client/pull/555): Make page size in resolving IDs configurable - [@eizengan](https://github.com/eizengan).
 * Your contribution here.
 
 ### 2.5.2 (2025/02/19)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * [#549](https://github.com/slack-ruby/slack-ruby-client/pull/549): Add group ID formatting support for message mentions - [@n0h0](https://github.com/n0h0).
 * [#553](https://github.com/slack-ruby/slack-ruby-client/pull/553): Use `secure_compare` during signature verification - [@hieuk09](https://github.com/hieuk09).
+* [#555](https://github.com/slack-ruby/slack-ruby-client/pull/555): Make page size in resolving conversation IDs configurable - [@eizengan](https://github.com/eizengan).
 * Your contribution here.
 
 ### 2.5.2 (2025/02/19)
--- a/lib/slack/web/api/mixins/conversations.id.rb
+++ b/lib/slack/web/api/mixins/conversations.id.rb
@@ -12,8 +12,12 @@ module Slack
           #
           # @option options [channel] :channel
           #   Channel to get ID for, prefixed with #.
+          # @option options [integer] :limit
+          #   The page size used for conversations_list calls required to find the channel's ID
           def conversations_id(options = {})
             name = options[:channel]
+            limit = options.fetch(:limit, conversations_id_page_size)
+
             raise ArgumentError, 'Required arguments :channel missing' if name.nil?
 
             id_for(
@@ -23,7 +27,7 @@ module Slack
               :conversations_list,
               :channels,
               'channel_not_found',
-              enum_method_options: { limit: conversations_id_page_size }
+              enum_method_options: { limit: limit }
             )
           end
         end
--- a/lib/slack/web/api/mixins/conversations.id.rb
+++ b/lib/slack/web/api/mixins/conversations.id.rb
@@ -12,22 +12,21 @@ module Slack
           #
           # @option options [channel] :channel
           #   Channel to get ID for, prefixed with #.
-          # @option options [integer] :limit
+          # @option options [integer] :id_limit
           #   The page size used for conversations_list calls required to find the channel's ID
           def conversations_id(options = {})
             name = options[:channel]
-            limit = options.fetch(:limit, conversations_id_page_size)
+            limit = options.fetch(:id_limit, conversations_id_page_size)
 
             raise ArgumentError, 'Required arguments :channel missing' if name.nil?
 
             id_for(
-              :channel,
-              name,
-              '#',
-              :conversations_list,
-              :channels,
-              'channel_not_found',
-              enum_method_options: { limit: limit }
+              key: :channel,
+              name: name,
+              prefix: '#',
+              enum_method: :conversations_list,
+              list_method: :channels,
+              options: { limit: limit }
             )
           end
         end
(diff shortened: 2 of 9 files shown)


eizengan:
Looks like this will require new cassettes. Presumably maintainers will want those made for workspaces to which they have access? I'll need someone else to record them, if so.

dblock (changes requested):
Thanks, this is a good change. However you should implement it in a way that does not change the existing functionality. 

I think we don't want to have to pass the limit all over the place, but to configure it globally and per/client, so follow the existing settings pattern to add it. The default should be no changes, making the API calls as today.

eizengan:
> I think we don't want to have to pass the limit all over the place

It's only passed in the one spot, and should probably remain there; if we defer until we're inside `id_for` we either:
- lose the context of this being a `conversations_id` call instead of e.g. `users_id`, which has a different maximum page size
- have to build an understanding of the caller inside `id_for`, (i.e. branch on `enum_method`)

> ...implement it in a way that does not change the existing functionality...
> ...configure it globally and per/client...
Amended, pushed. How do we feel about the name `conversations_id_page_size`?

The config change does create an unusual codependency between `Slack::Web::Api::Mixins::Conversations` and `Slack::Web::Config`, as illustrated by [this failing test run](https://github.com/slack-ruby/slack-ruby-client/actions/runs/14756788400/job/41427000308?pr=555). The second commit here provides one possible solution.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 * [#549](https://github.com/slack-ruby/slack-ruby-client/pull/549): Add group ID formatting support for message mentions - [@n0h0](https://github.com/n0h0).
 * [#553](https://github.com/slack-ruby/slack-ruby-client/pull/553): Use `secure_compare` during signature verification - [@hieuk09](https://github.com/hieuk09).
+* [#555](https://github.com/slack-ruby/slack-ruby-client/pull/555): Make page size while looking for a Conversation's ID configurable - [@eizengan](https://github.com/eizengan).
```
dblock, line 5: Let's at least lowercase conversation.

```suggestion
* [#555](https://github.com/slack-ruby/slack-ruby-client/pull/555): Make page size in resolving conversation IDs configurable - [@eizengan](https://github.com/eizengan).
```

or

```suggestion
* [#555](https://github.com/slack-ruby/slack-ruby-client/pull/555): Added `config.conversations_id_page_size` - [@eizengan](https://github.com/eizengan).
```

dblock (changes requested):
We are _still_ changing behavior by passing in a page size (100) by default. The code should not be passing anything into `limit` as before by default to avoid breaking backwards compatibility. I understand this may be the Slack default, or we have `default_page_size`, but we are relying on this knowledge here.  

Stepping back, I propose we add `id_for_page_size` instead of `conversations_id_page_size` and implement this inside `id_for`. It will be simpler and cleaner. Unless you have a reason why this may be a bad idea?

And I'd like to be able to write this, overriding whatever the global `id_for_page_size` was set to (needs tests):

```ruby
conversations_id(channel: 'C1234', limit: 10)
```

Sorry to be a pest, hang in here with me :)

eizengan:
> Stepping back, I propose we add id_for_page_size instead of conversations_id_page_size and implement this inside id_for. It will be simpler and cleaner. Unless you have a reason why this may be a bad idea?

My only major concern is hypothetical: there might be scenarios where no single page size is good for all translations using `id_for`. I think we only use this to translate users and channels at present, so my concern may be unwarranted. We can easily pivot to a single `id_for_page_size` config value if desired.

> We are still changing behavior by passing in a page size (100) by default.

I think you refer to theh explicit 100 I used to appease the test suite? If so, I'd overlooked the fact we were already doing some stubbing there. Oops! Switched to a stub - no more 100.

> And I'd like to be able to write this, overriding whatever the global id_for_page_size was set to (needs tests):
>
> `conversations_id(channel: 'C1234', limit: 10)`

I like it! Coming shortly...

eizengan:
I've added a couple commits to implement the requested `limit` parameter and to extend configuration to `users_id`. We can instead consolidate those config options into a single `id_for_page_size` quite easily if that's your preference.

lib/slack/web/api/mixins/conversations.id.rb
```diff
@@ -12,11 +12,23 @@ module Conversations
...
 
-            id_for :channel, name, '#', :conversations_list, :channels, 'channel_not_found'
+            id_for(
+              :channel,
+              name,
+              '#',
+              :conversations_list,
+              :channels,
+              'channel_not_found',
+              enum_method_options: { limit: limit }
```
dblock, line 30: I think either we just add one more `limit` parameter to the already long list of arguments or change the signature to just `id_for(options)`? WDYT?

If you think both are a bad idea, at a minimum let's rename `enum_method_options` to just `options` to be future-proof?

dblock (changes requested):
Much better!

Add tests for the default value and I think we need to rename things.

eizengan:
> Add tests for the default value

Do the changes in the first new commit suffice? I'm not 100% sure on where these should go otherwise. Tests for page size are not present near the other config value tests.

> I think either we just add one more limit parameter to the already long list of arguments or change the signature to just id_for(options)? WDYT?

While considering this I noticed a Problem; these methods are often used inside [endpoint methods](https://github.com/slack-ruby/slack-ruby-client/blob/master/lib/slack/web/api/endpoints/conversations.rb#L122) which - as a result of our change - unintentionally pass `limit` arguments intended for the parent call. I've tentatively renamed our new parameter `id_limit` to avoid this.

A single `options` parameter feels fragile to me since it would need exactly the right keys to work correctly. IMO keyword arguments feel like a good choice here.

The second new commit contains changes related to the above issues.

spec/slack/web/api/mixins/conversations_spec.rb
```diff
@@ -31,15 +31,39 @@
     end
 
     it 'translates a channel that starts with a #' do
+      expect(conversations).to receive(:conversations_list).with(limit: 100)
```
dblock, line 34: This is a change in default behavior isn't it? This test doesn't pass without the changes.

lib/slack/web/api/mixins/ids.id.rb
```diff
@@ -6,15 +6,16 @@ module Mixins
...
             return { 'ok' => true, key.to_s => { 'id' => name } } unless name[0] == prefix
 
-            public_send enum_method do |list|
+            public_send(enum_method, **options) do |list|
               list.public_send(list_method).each do |li|
                 return Slack::Messages::Message.new('ok' => true, key.to_s => { 'id' => li.id }) if li.name == name[1..-1]
               end
             end
 
+            not_found_error = "#{key}_not_found"
```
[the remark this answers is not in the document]
eizengan: Fair point!

spec/slack/web/api/mixins/conversations_spec.rb
eizengan: Ah, yes - the old code relies on the default from within the paginator. `Hash#compact` should suffice to fix this if the defaults are nil; see what you think

eizengan:
Something funky going on in tests - lots of
```
NameError:
  undefined method 'parse' for class '#<Class:CGI>'
```
followed by a segfault. Don't have time to dig in at the moment; if you do and hit paydirt, please let me know!

eizengan:
Modified your CI so e.g. when ruby-head segfaults during tests Coveralls doesn't panic over the missing test info; ready for re-review

eizengan:
> Thanks for hanging in here with me

Not at all; thanks for ensuring additions to this gem are appropriately refined!

PR #559, opened 2025-06-24
Enable name-to-id translation for private channels

eizengan:
Ran into another hitch with name-to-id translation: the gem is only capable of translating public channels.

This first pass does the minimum amount of work to make the feature happen. It has an obvious flaw, however, in that the continued addition of [new arguments](https://api.slack.com/methods/conversations.list#args) feels untenable (e.g. team_id or exclude_archived). Perhaps this is the last argument we need/want, and the first-pass solution is fine - or perhaps not. It's a good starting point for conversation, regardless. We can refine once I hear your thoughts.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,6 +30,7 @@ jobs:
         with:
           ruby-version: ${{ matrix.entry.ruby }}
           bundler-cache: true # 'bundle install' and cache gems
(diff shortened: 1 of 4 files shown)


dblock:
The number of parameters passed here is limited, so I am ok with this. Want to open a PR for the rest of them?

eizengan:
Closing the loop: the other additions are in #560

PR #564, opened 2025-07-11
Raise InvalidSignature if Slack::Events::Request has no signature

### LINKED ISSUE #563: SlackRequest verify raises `undefined method `bytesize' for nil` if missing signature
Minor regression after https://github.com/slack-ruby/slack-ruby-client/pull/553

If you instantiate a request without a signature, instead of getting `InvalidSignature` when calling `verify` you get `undefined method`

```
=> #<NoMethodError: undefined method `bytesize' for nil>
[2] pry(#<Integrations::Slack::WebhooksController>)> error.backtrace
=> ["/Users/wesley/.asdf/installs/ruby/3.3.7/lib/ruby/gems/3.3.0/gems/slack-ruby-client-2.6.0/lib/slack/utils/security.rb:38:in `secure_compare'",
 "/Users/wesley/.asdf/installs/ruby/3.3.7/lib/ruby/gems/3.3.0/gems/slack-ruby-client-2.6.0/lib/slack/events/request.rb:63:in `valid?'",
...
```

wesleyjellis:
Fix regression after #553. The old implementation handled a nil value properly, but the new implementation raises as `nil` does not have `bytesize`

Closes #563 

(Will fix the changelog in a second)


dblock:
Thanks for fixing this!

PR #583, opened 2026-02-19
Not found errors raised by id_for use more specific error classes when they exist

eizengan:
### Problem

Presently the `id_for` method always raises a `SlackError`. This means that when calling e.g. `users_info` one must handle both `UserNotFound` and `SlackError`-with-a-message-equal-"user_not_found" to actually cover ones bases. 

### Solution

This uses the existing error type deduction mechanism to raise a more specific error when one exists. At the time of writing this will work for all consumers of `id_for`

### Potential issues

The specific errors are subclasses of `SlackError`, and therefore will still behave as expected in existing `rescue`s, `is_a?` calls, etc. Any class equality checks failing to account for inheritance, however, will break.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-### 3.1.1 (Next)
+### 3.2.0 (Next)
 
 * [#581](https://github.com/slack-ruby/slack-ruby-client/pull/581): Migrate Danger to use danger-pr-comment workflow - [@dblock](https://github.com/dblock).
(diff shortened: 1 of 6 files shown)


mformus (approved):

dblock (changes requested):
LGTM, thank you! Let's add a section to UPGRADING and bump the version to 3.2.0 since it's not a minor change. 

It looks backwards compatible since `NotFound` errors are also `SlackError`'s? Check?

dblock (dismissed):

